Issue #239, opened 2014-06-04
uniform distribution sampler returns inf


rebeccaem:
Recreate:

```c++
int main (void) {
  QUESO::VectorSpace<QUESO::GslVector,QUESO::GslMatrix>
    paramSpace(env, "param_", 1, NULL);

  QUESO::GslVector paramMins(paramSpace.zeroVector());
  paramMins.cwSet(0.);
  QUESO::GslVector paramMaxs(paramSpace.zeroVector());
  paramMaxs.cwSet(INFINITY);
  QUESO::BoxSubset<QUESO::GslVector,QUESO::GslMatrix>
    paramDomain("param_", paramSpace, paramMins, paramMaxs);

  QUESO::VectorSpace<QUESO::GslVector,QUESO::GslMatrix>
    qoiSpace(env, "qoi_", 1, NULL);

  QUESO::UniformVectorRV<QUESO::GslVector,QUESO::GslMatrix>
    paramRv("param_", paramDomain);

  QUESO::GslVector v(paramSpace.zeroVector());
  for (unsigned int i = 0; i < 10; i++) {
    paramRv.realizer().realization(v);
    std::cout << "sample is: " << v[0] << std::endl;
  }
  return 0;
}
```

Produces the following output:

```
sample is: inf
sample is: inf
sample is: inf
sample is: inf
sample is: inf
sample is: inf
sample is: inf
sample is: inf
sample is: inf
sample is: inf
```

Replacing `INFINITY` with `5.0` gives expected results.

halp

dmcdougall:
Fixed in #259.

PR #259, opened 2014-08-14
Add doc notes RE draws from an improper uniform

dmcdougall:
Addresses #239.


dmcdougall:
I've compromised on throwing an exception from the realizer, not the constructor.  That way users code will still work with an improper uniform for the inverse problem, but will error in the forward problem.  Before this change, rather than an error being thrown from the forward problem, `inf` was drawn from the `UniformVectorRealizer`.  At least now we know why it's `inf`.

PR #286, opened 2014-11-11
New libmesh compatibility

roystgnr:
This gets me building against libMesh master (as of a few minutes ago when we restored the libMeshEnums namespace https://github.com/libMesh/libmesh/pull/407 ) but "make check" is dying with diffs in test/t01_valid_cycle


roystgnr:
Replaced by #339

PR #306, opened 2015-01-06
Fix sign in log uniform pdf

dmcdougall:
Queso was returning the wrong value for log of a uniform pdf.  While this doesn't change the solution for sampling or minimisation, it was very confusing when debugging when I observed a sign change in my objective function that shouldn't have happened.  `actualValue` returns the correct value, only `lnValue` was incorrect.


nicholasmalaya:
Looks good to me. Simple fix, nice find.

PR #310, opened 2015-01-21
Add input file options for toggling optimisation and monitoring

dmcdougall:


src/stats/src/StatisticalInverseProblem.C
```diff
@@ -256,11 +257,25 @@ StatisticalInverseProblem<P_V,P_M>::solveWithBayesMetropolisHastings(
   // user-provided initial seed, or use the user-provided seed for a
   // deterministic optimisation instead and seed the chain with the result of
   // the optimisation
-  if (this->m_seedWithMAPEstimator) {
+  if (this->m_seedWithMAPEstimator ||
+      m_optionsObj->m_ov.m_seedWithMAPEstimator) {
+    // Unfortunately, I didn't have the foresight to make this an input file
+    // option from the beginning, hence needing to check two flags
```
pbauman: Why do you still need to keep both around?

pbauman:
Otherwise :+1: after rebase.

src/stats/src/StatisticalInverseProblem.C
dmcdougall: Backwards compatibility.

dmcdougall:
I've updated this to use the new options parsing foo.

pbauman:
Source changes look OK. Looks like 448d56d failed `make check` though...

dmcdougall:
That should be better.

PR #312, opened 2015-01-23
Implement canned Gaussian likelihoods

dmcdougall:
This PR implements classes that represent Gaussian-form likelihoods.  Note that, if the user's model is not linear, the likelihood is not actually Gaussian, but the expression for the probability distribution function implies a Gaussian distribution for the observational error on the observations.

An example is provided, and a test will be added later today.  I wanted to get this up to elicit any initial feedback people might have on the design.  Looking at the example is a good first start.

### DIFF
--- a/src/core/inc/GslBlockMatrix.h
+++ b/src/core/inc/GslBlockMatrix.h
@@ -101,7 +101,6 @@ public:
   //@}
 
 private:
-  const BaseEnvironment & m_env;
   std::vector<VectorSpace<GslVector, GslMatrix> *> m_vectorSpaces;
   std::vector<GslMatrix *> m_blocks;
 };
--- a/src/core/inc/GslBlockMatrix.h
+++ b/src/core/inc/GslBlockMatrix.h
@@ -30,7 +30,11 @@
  * \brief QUESO matrix class using GSL.
  */
 
-#include <queso/Matrix.h>
+#include <vector>
+#include <queso/Environment.h>
+#include <queso/GslVector.h>
+#include <queso/GslMatrix.h>
+#include <queso/VectorSpace.h>
 
 namespace QUESO {
 
@@ -43,21 +47,21 @@ namespace QUESO {
  * by an encapsulated gsl_matrix structure.
  */
 
-class GslBlockMatrix : public Matrix
+class GslBlockMatrix
 {
 public:
   //! @name Constructor/Destructor methods
   //@{
   //! Shaped Constructor: creates a square matrix with size \c v.sizeLocal() and diagonal values all equal to \c diagValue.
-  GslBlockMatrix(const std::vector<unsigned int> & blockSizes,
-      double diagValue);
+  GslBlockMatrix(const FullEnvironment & env,
+      const std::vector<unsigned int> & blockSizes, double diagValue);
 
   //! Destructor
   ~GslBlockMatrix();
   //@}
 
   //! Return block \c i in the block diagonal matrix
-  GslBlockMatrix & getBlock(unsigned int i) const;
+  GslMatrix & getBlock(unsigned int i) const;
 
   //! Return the number of blocks in the block diagonal matrix
   unsigned int numBlocks() const;
@@ -67,7 +71,7 @@ public:
    * It checks for a previous LU decomposition of \c this matrix and does not
    * recompute it if m_MU != NULL.
    */
-  void invertMultiply (const GslVector & b, GslVector & x) const;
+  void invertMultiply(const GslVector & b, GslVector & x) const;
 
   //! @name I/O methods
   //@{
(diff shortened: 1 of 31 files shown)


dmcdougall:
The example is [here](https://github.com/dmcdougall/queso/blob/dakota-canned-llhd/examples/gaussian_likelihoods/scalarCovariance.C).

dmcdougall:
@pbauman Look at the example here, and tell me if you think this is a reasonable workflow.

pbauman:
Just glanced at the example, not the underlying classes - looks awesome. I would expect something exactly like you have. I'm excited to be able to leverage this.

dmcdougall:
@pbauman Check out the [diagonal](https://github.com/dmcdougall/queso/blob/dakota-canned-llhd/examples/gaussian_likelihoods/diagonalCovariance.C) and [full](https://github.com/dmcdougall/queso/blob/dakota-canned-llhd/examples/gaussian_likelihoods/fullCovariance.C) examples.  Same vein.

pbauman:
:+1:

dmcdougall:
@pbauman And, last one, the [block diagonal case](https://github.com/dmcdougall/queso/blob/dakota-canned-llhd/examples/gaussian_likelihoods/diagonalCovariance.C) is ready, too.

The only things left to do are to update documentation and write tests.  I can do that over the weekend.  This is almost complete now.

dmcdougall:
Tests are in.  Will comb over the docs now to make sure everything is clear.

dmcdougall:
Documentation is tidied up.

This is ready to go pending feedback from the DAKOTA folks and possible @pbauman as well.

src/core/inc/GslBlockMatrix.h
```diff
@@ -0,0 +1,94 @@
...
+
+/*!
+ * \class GslBlockMatrix
+ * \brief Class for representing block matrices using GSL library.
+ *
+ * This class provides basic 'invertMultiply' support for matrices of block
+ * diagonal structure.  Each block is implemented as a GslMatrix object.
+ */
+
+class GslBlockMatrix
```
pbauman: You don't subclass Matrix?

src/stats/src/GaussianLikelihoodBlockDiagonalCovariance.C
```diff
@@ -0,0 +1,93 @@
...
+    m_covariance(covariance)
+{
+  unsigned int totalDim = 0;
+
+  for (unsigned int i = 0; i < this->m_covariance.numBlocks(); i++) {
+    totalDim += this->m_covariance.getBlock(i).numRowsLocal();
+  }
+
+  if (totalDim != observations.sizeLocal()) {
+    queso_error_msg("Covariance matrix not same size as observation vector");
```
pbauman: Would say "dimension" instead of "size"

src/core/src/GslBlockMatrix.C
```diff
@@ -0,0 +1,122 @@
...
+
+  if (x.sizeLocal() != b.sizeLocal()) {
+    queso_error_msg("solution and rhs have incompatible sizes");
+  }
+
+  unsigned int blockOffset = 0;
+
+  // Do an invertMultiply for each block
+  for (unsigned int i = 0; i < this->m_blocks.size(); i++) {
+    GslVector blockRHS(this->m_vectorSpaces[i]->zeroVector());
```
pbauman: If this pops up in profiling, it might be worth implementing a version of invertMultiply that populates a solution vector that you give to it so we don't have to copy back and forth.

pbauman:
@dmcdougall What's the status on this? I may be able to help chip in in a few days if it would help close this out.

dmcdougall:
I'm going to write a few more tests for `GslBlockMatrix`, and address some of your comments, and then this is good to go.  I'm going to base some more canned likelihoods (for the DAKOTA folks) off of this branch.  Will tidy up this branch now though, so you can play asap.

dmcdougall:
Oh I should also update the changelog.  I always forget to do that.

src/core/inc/GslBlockMatrix.h
dmcdougall: Nope.  Do you think this should?  I thought composition was enough here, but I can make it a subclass if you think that's a better design.

dmcdougall:
Ok, this is good to go I think.  I also checked each commit passed `make check`.

@pbauman Give this a once-over?

src/core/inc/GslBlockMatrix.h
pbauman: http://i.imgur.com/OrUWTpY.jpg

It's a matrix class used by QUESO. Seems like it should adhere to QUESO's Matrix interface. You can still use GSLMatrix for the composition aspects, but it should subclass QUESO::Matrix to adhere to the interface and so we can use it as such.

pbauman: That said, I wouldn't object to having the missing pure virtual methods throw a queso_not_implemented() so this can move forward.

dmcdougall: >That said, I wouldn't object to having the missing pure virtual methods throw a queso_not_implemented() so this can move forward.

Thanks for the input.  I've pushed a commit that addresses this.

dmcdougall:
@pbauman Is e6b7ec0 kosher?

pbauman:
Barring that one comment, this looks good.

dmcdougall:
@pbauman 9ffbab5 addresses the duplicated member variable.

pbauman:
:+1:

Issue #308, opened 2015-01-15
UQ TEST FATAL ERROR, rank 2, in SequenceOfVectorss<V,M>::getPositionValues(): posId is NULL. Exiting...


dmcdougall:
This file produces the error on 4 processors, but not one:

```
#include <queso/GslVector.h>
#include <queso/GslMatrix.h>
#include <queso/UniformVectorRV.h>
#include <queso/StatisticalInverseProblem.h>
#include <queso/ScalarFunction.h>
#include <queso/VectorSet.h>

template<class V, class M>
class Likelihood : public QUESO::BaseScalarFunction<V, M>
{
public:

  Likelihood(const char * prefix, const QUESO::VectorSet<V, M> & domain)
    : QUESO::BaseScalarFunction<V, M>(prefix, domain)
  {
  }

  virtual ~Likelihood()
  {
  }


  virtual double lnValue(const V & domainVector, const V * domainDirection,
      V * gradVector, M * hessianMatrix, V * hessianEffect) const
  {
    return -0.5 * domainVector[0] * domainVector[0];
  }

  virtual double actualValue(const V & domainVector, const V * domainDirection,
      V * gradVector, M * hessianMatrix, V * hessianEffect) const
  {
    return std::exp(this->lnValue(domainVector, domainDirection, gradVector,
          hessianMatrix, hessianEffect));
  }
};

int main(int argc, char ** argv) {
  MPI_Init(&argc, &argv);

  unsigned int dim = 1;

  QUESO::FullEnvironment env(MPI_COMM_WORLD, argv[1], "", NULL);

  QUESO::VectorSpace<QUESO::GslVector, QUESO::GslMatrix> paramSpace(env,
      "param_", dim, NULL);

  double min_val = -1.0;
  double max_val = 1.0;

  QUESO::GslVector paramMins(paramSpace.zeroVector());
  paramMins.cwSet(min_val);
  QUESO::GslVector paramMaxs(paramSpace.zeroVector());
  paramMaxs.cwSet(max_val);

  QUESO::BoxSubset<QUESO::GslVector, QUESO::GslMatrix> paramDomain("param_",
      paramSpace, paramMins, paramMaxs);

  QUESO::UniformVectorRV<QUESO::GslVector, QUESO::GslMatrix> priorRv("prior_",
      paramDomain);

  Likelihood<QUESO::GslVector, QUESO::GslMatrix> lhood("llhd_", paramDomain);

[...]

PR #314, opened 2015-02-05
Add test for BaseVectorSequence::unifiedPositionsOfMaximum

dmcdougall:


dmcdougall:
This should now fix #308.  @beanSnippet can you confirm this?  And also that `make check` passes, too?

dmcdougall:
I think #308 is serious enough to warrant a `v0.51.1` bugfix release.

@beanSnippet As soon as you let me know this fixes the problem, I can get that sorted out.

beanSnippet:
Make check looks good. With my code however, sometimes it works on multiple
processors and sometimes it doesn't! Just changing the number of samples in
the chain can be the difference. So I'm not sure what's going on now...

dmcdougall:
>Make check looks good. With my code however, sometimes it works on multiple
processors and sometimes it doesn't! Just changing the number of samples in
the chain can be the difference. So I'm not sure what's going on now...

With the same example code?  And are you getting the same error message?

beanSnippet:
Yes, same example code and same error message. The only difference I've found so far is that it will work when the chain length is <5 (before it was always failing on multiple processors), but not for >= 5.

beanSnippet:
Was a path problem on my end, the fix holds, thanks @dmcdougall (for the fix and path help).

dmcdougall:
Happy to help.

I'll merge this into a bugfix branch first so I can roll up a bugfix release.  That way the DAKOTA folks don't get bitten.

dmcdougall:
Replaced by #324.

PR #319, opened 2015-03-06
Add a coefficient for cov matrix in Gaussian llhd

dmcdougall:
The user can specify a multiplicative coefficient for the covariance matrix in a Gaussian likelihood.

### DIFF
--- a/examples/gaussian_likelihoods/fullCovariance.C
+++ b/examples/gaussian_likelihoods/fullCovariance.C
@@ -35,12 +35,12 @@ template<class V, class M>
 class Likelihood : public QUESO::GaussianLikelihoodFullCovariance<V, M>
 {
 public:
-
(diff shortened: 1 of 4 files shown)


dmcdougall:
@pbauman A slight update to the full covariance matrix case merged in #312.  Does this look good to you?  Another approach would be to make the coefficient have a default value of 1.0 to keep clutter down.  Thoughts?

pbauman:
Yeah, I'd default that to 1.0. What's the use case out of curiosity?

dmcdougall:
It's for a DAKOTA FY15 milestone.

dmcdougall:
@pbauman Defaulting the value made things a lot cleaner.  Take a look.  `make check` passes.

pbauman:
:+1:

PR #321, opened 2015-03-06
Adding canned Gaussian likelihood with random multiplicative hyperparameter

dmcdougall:
This pull request implements a Gaussian likelihood where the multiplicative factor in front of the observation error covariance matrix is treated as a random variable.

The user must specify the appropriate dimension in the vector space for the `StatisticalInverseProblem` class, but that is all.  The user does not have to deal with the hyperparameter explicitly in the model evaluation, since this is dealt with directly in the likelihood.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -175,3 +175,4 @@ test/test_blockDiagonalCovariance
 test/test_output_gaussian_likelihoods
 test/test_gaussian_likelihoods/queso_input.txt
 test/test_GslBlockMatrixInvertMultiply
+test/test_fullCovarianceRandomCoefficient
--- a/.gitignore
+++ b/.gitignore
@@ -87,6 +87,7 @@ examples/scalarCovariance
 examples/diagonalCovariance
 examples/fullCovariance
 examples/blockDiagonalCovariance
+examples/fullCovarianceRandomCoefficient
 
 src/core/inc/queso.h
 src/libqueso.la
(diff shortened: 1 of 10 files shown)


dmcdougall:
@pbauman It seems like you're my guinea pig today.  Fancy looking this over?  Each commit passes `make check`.

pbauman:
Just scanning through the source looks good. I could tell what's going on from the documentation.

pbauman:
Actually, a thought just occurred to me that will apply to all these PRs.

We should only pass the part of the domainVector that the user cares about, i.e. do not pass them the part of the vector that has the hyperparameters. This removes the chance of the user inadvertently mucking with it.

dmcdougall:
>We should only pass the part of the domainVector that the user cares about, i.e. do not pass them the part of the vector that has the hyperparameters. This removes the chance of the user inadvertently mucking with it.

The `domainVector` variable is of type `const V &`, so the user can't change it before queso gets it back to evaluate the likelihood.  The only concern is that by the giving the user access to it, the user could inadvertently use it in their model evaluation.  This would be a bug in the user's code.  I'm in favour of design to prevent user error, but in this particular situation the user has to [set up the domain to contain the hyperparameters anyway](https://github.com/libqueso/queso/pull/321/files#diff-2492b64da70e1e42879cad3927f37f69R85).


I'm tempted to pass the whole thing for now;  I'd rather have the consistency between the dimension of `domainVector` and the dimension of the `VectorSpace` object the user sets up to house the inverse problem.

dmcdougall:
@pbauman Are you happy with my rebuttal?  If so, shall I merge this?

pbauman:
Yeah. I think since the hyperparameter is part of the class, the user should be excused from the burden of managing it, but I can see it both ways. Go for it.

dmcdougall:
Done.

PR #323, opened 2015-03-07
Add multiplicative coefficients to canned Gaussian likelihood with block diagonal covariance matrix

dmcdougall:
This PR adds multiplicative coefficients for each block in the block diagonal covariance matrix for the canned Gaussian likelihood.  The default value for the coefficients is 1.0.

The test has been updated to use the coefficients, and `make check` passes.


dmcdougall:
@pbauman Happy with this design?

pbauman:
These aren't hyperparameters automatically, right? If that statement is correct, then :+1:

dmcdougall:
Correct.  Merging.

PR #324, opened 2015-03-09
Fix unified positions of maximum

dmcdougall:
Replaces #314.


dmcdougall:
Each commit passes `make check`.  Merging.

PR #325, opened 2015-03-10
Add accessors for chain, log-likelihood and log-target sequences.

briadam:
Allow C++ clients access to stored sequences for use in
post-processing and output.


briadam:
While it would help our sync workflow to also have this patched onto the current v0.51.1 release (or even earlier as we first starting applying it to v0.50.0), it's more of an enhancement, so I realize that might not make sense.

dmcdougall:
This looks fine to me.  Thanks for the patch @briadam, and thanks for adding doxygen comments.

PR #330, opened 2015-03-18
Fix constructors that throw errors

### LINKED ISSUE #211: Why do constructors exist to fail
See [here](https://github.com/libqueso/queso/blob/8c10d13cc4906364e992bcfc098bea6865887aec/src/core/src/Vector.C#L37).  That constructor always fails.  Is there a way of calling that constructor without it failing?  If not, then this and all other constructors that exist to fail should just be removed.

dmcdougall:
Fixes #211.

Default constructors are made `private`, and copy constructors are removed.


dmcdougall:
@pbauman This look ok?

pbauman:
:+1:

dmcdougall:
And just to clarify, the removal of copy constructors breaks backwards compatibility in the cases where the the compiler does not create one for you.  Should we take extra provisions to deal with this?

pbauman:
Are there any cases where it *wasn't* meant to error? I thought all the cases were instances of immediately throw error so, in fact, we're helping out the user.

dmcdougall:
No.  In which case I'll merge this.  Thanks.

dmcdougall:
That merge was painful.

PR #333, opened 2015-03-24
Add explicit cast with template parameters for GPMSA.

briadam:
This resolves ambiguity that gave gcc-4.3.2 trouble.

Comes from Dakota upstream: https://software.sandia.gov/trac/tpl/changeset/2747


dmcdougall:
Thanks!

PR #337, opened 2015-04-07
Fix logit transformation of proposal covariance matrix

dmcdougall:
Fixes #316.

PR #338, opened 2015-04-07
Fix logit transformation of proposal covariance matrix

dmcdougall:
This replaces #337 but targets the bugfix branch instead of dev.

@beanSnippet Do you mind quickly giving this a test drive?


beanSnippet:
This is looking good. I can still force a nan with an unreasonable proposal covariance, but now I see the error message:
```
Proposal variance element 1 is 50 but domain is of size 3
QUESO does not support uniform-like proposal distributions.  Try making the proposal variance smaller
```
Anything else to test?

dmcdougall:
@beanSnippet Nope.  Merging.

PR #342, opened 2015-04-15
Implement default template parameters

dmcdougall:
I decided to have a go at setting `GslVector` and `GslMatrix` as default template parameters so the user didn't have to do so much typing to create instantiations of `QUESO`'s classes.  Everything compiles and the tests pass locally and on Travis.

While not necessary, it might be a good idea to refactor the examples to utilise these to encourage new users to take advantage of this new pattern.  I think this is also serves as a good middle ground between now and removing the templates entirely, thus breaking backwards compatibility.

### DIFF
--- a/examples/gravity/src/gravity_compute.C
+++ b/examples/gravity/src/gravity_compute.C
@@ -88,8 +88,7 @@ void computeGravityAndTraveledDistance(const QUESO::FullEnvironment& env) {
   //------------------------------------------------------
   // SIP Step 1 of 6: Instantiate the parameter space
   //------------------------------------------------------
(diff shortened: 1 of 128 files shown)


dmcdougall:
@beanSnippet We talked the other day about this.  You can try it out and let me know what you think before it gets merged.

beanSnippet:
Looks good to me!

dmcdougall:
@pbauman Thoughts on this?

pbauman:
Source looks fine. (We'll need to make sure to update the new stuff I added/am adding accordingly, depending on who merges first.) Only other comment would be to update one of the examples to illustrate use and comment and one of the tests to capture any compilation problems.

dmcdougall:
@pbauman Done and done.  When Travis comes back happy I'll ship this.

Issue #341, opened 2015-04-09
Remove support for discrete distributions?


dmcdougall:
There are a couple of classes that look like an initial attempt at trying to support discrete distributions and vector spaces.  Namely [FiniteDistribution](https://github.com/libqueso/queso/blob/dev/src/stats/inc/FiniteDistribution.h) and [DiscreteSubset](https://github.com/libqueso/queso/blob/dev/src/basic/inc/DiscreteSubset.h).

I vote to remove these classes.  None of the examples use them.

dmcdougall:
These are now deprecated, and should be removed after the next release is out.

dmcdougall:
Addressed in #562.

PR #343, opened 2015-04-16
Deprecate `FiniteDistribution` and `DiscreteSubset`

dmcdougall:
Addresses first stage of #341.


dmcdougall:
Anybody fancy doing a quick review of this?

pbauman:
Was this functional? If not, then :+1: If so, do you think it's worth keeping?

dmcdougall:
> Was this functional?

I have no idea; we have no tests for it.

> If so, do you think it's worth keeping?

Supporting discrete random variables?  No.  The transition kernel looks totally different.

dmcdougall:
@pbauman Thanks for the review.  I'm going to merge this.  If anybody has any objections we can revert the commit, otherwise I think these classes should eat it.

PR #344, opened 2015-04-16
Deprecate `UQ_*` macros

### LINKED ISSUE #74: Deprecate MACROS in Define.h
The Macros in src/core/inc/Defines.h:

* UQ_RC_MACRO
* UQ_TEST_MACRO
* UQ_FATAL_RC_MACRO
* UQ_FATAL_TEST_MACRO

Are all old and do not appropriately understand MPI. Let's replace them with queso_exit (and associated) calls, which can provide catchable error handling as well as better information on what caused the failures.

dmcdougall:
Fixes #74.

Note:  #343 should be merged before this (that's where I added the `queso_deprecated()` macro).

### DIFF
--- a/src/basic/src/DiscreteSubset.C
+++ b/src/basic/src/DiscreteSubset.C
@@ -35,6 +35,8 @@ DiscreteSubset<V,M>::DiscreteSubset(const char* prefix,
   : VectorSubset<V,M>(prefix, vectorSpace, 0.),
     m_elements(elements.size(),NULL)
 {
+  queso_deprecated();
+
   m_volume = 0.;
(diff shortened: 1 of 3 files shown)


roystgnr:
The non-deprecated/preferred calls are queso_error() and kin?

pbauman:
One thing we might want to do too is instead of a `--enable-debug` option is a `--enable-opt` option to throw a `-DNDEBUG` into the compilation flags since the `queso_assert()` etc. will be enabled until one throws that flag in.

dmcdougall:
> The non-deprecated/preferred calls are queso_error() and kin?

Yes.

Issue #33, opened 2013-09-12
Make fortran binding example build


dmcdougall:
See `examples/statisticalInverse_Fortran`

dmcdougall:
Fixed in #345.

PR #345, opened 2015-04-17
Fix fortran example

dmcdougall:
Partially addresses #33.

While #33 suggests making it build, building this with our build system will mean supporting all the different fortran/C++ linking idiosyncrasies between all the different fortran compilers.  Instead of this maintenance headache, I've updated the example to use the new queso api (new uq* prefix) and I've left the example script in there that shows how one might go about compiling and linking to queso.

I think that's the best course of action here.


dmcdougall:
I'm merging this, since it doesn't affect any of the testing or algorithms.

Issue #346, opened 2015-04-17
Refactor input file options handling


dmcdougall:
This is probably best explained with an example.

To specify input file options for the `StatisticalInverseProblem` class, one can [pass a pointer to an object of type `SipOptionsValues` to the constructor](http://libqueso.com/queso/docs/html/a00151.html#a98ad98a7030b2c6577a0e840506bf74d).  This can be used instead of reading in options from an input file.  The user can set the members of `SipOptionsValues` (`m_computeSolution`, for example) for customisation.

However, `StatisticalInverseProblem` actually creates an entirely different object representing input file options (in this case, [`StatisticalInverseProblemOptions`](http://libqueso.com/queso/docs/html/a00152.html)), using `SipOptionsValues`, to actually read the input options (whether reading from an input file or not).  See [here](https://github.com/libqueso/queso/blob/dev/src/stats/src/StatisticalInverseProblem.C#L72) for where that happens.  Also notice that only the `StatisticalInverseProblemOptions` class can read from the input file, and the function [`scanOptionsValues()`](https://github.com/libqueso/queso/blob/dev/src/stats/src/StatisticalInverseProblem.C#L77) is the method that executes this.

This is the case for every object whose behaviour can be modified by input file options.

To reiterate, for every class whose behaviour is tweakable via an input file, we have __two more classes__ that store input file options and only one of which reads the input file.  Moreover, the classes that have a `scanOptionsValues()` method don't inherit from a base class that enforces this pattern.

Lastly, notice that `SipOptionsValues`, which can't read the input file, has all the options declared with usable types (`int`, `bool`, etc.) whilst `StatisticalInverseProblemOptions` has all the same options, but all of type `std::string`.  That is, they are useless to the object being tweaked by the input file.

I think it's time to clean this up.

I propose __one__ class that does both a) hold input options; and b) reads the input file, if necessary.  Additionally, this class would not store the useless `std::string` types but only the usable options.  This class would subclass a base class (maybe called `BaseInputFileOptions`) that contains a pure virtual function called `scanOptionsValues`, and any other methods common all the input file options classes currently in QUESO, namely [`defineMyOptions`](http://libqueso.com/queso/docs/html/a00152.html#ae085514e9153c131266f93abea2db3dd) and [`getMyOptionsValues`](http://libqueso.com/queso/docs/html/a00152.html#a5984ab1857cbe72f9f665388f9e87e5e).

Any objections?  Ideas?

dmcdougall:
[...]

PR #347, opened 2015-04-23
Refactor inputs to StatisticalInverseProblem

dmcdougall:
This PR implements what I think (with advice from @pbauman) is a good start for #346.  I've only done it for the `StatisticalInverseProblem` class because there aren't too many options for that so it's easier to see what's going on.

To summarise, I've deprecated all the methods in `StatisticalInverseProblemOptions`, which was created internally by `StatisticalInverseProblem`.  So we're not breaking backwards compatibility here.  I've moved all the parsing functionality into `SipOptionsValues`, which is the type of object that the user passes to the `StatisticalInverseProblem` constructor for custom options set programmatically.  All the options that were in `SipOptionsValues` have not moved, so we're not breaking backwards compatibility here either.  The `StatisticalInverseProblem` class now creates a `SipOptionsValues` object internally instead of a `StatisticalInverseProblemOptions`.  Lastly, I factored out the `scanOptionsValues` method into a `BaseInputOptions` class, because this method is common for all the other input option classes (which I haven't touched in this PR).  Later down the road, we'll get some code re-use out of that method.

Note that the class hierarchy is only one level.  The base class does the `boost`-specific parsing.  One could envisage inserting class in between `BaseInputOptions` and `SipOptionsValues` called `BaseBoostInputOptions` and making the `BaseBoostInputOptions` do all the boost-specific parsing instead.  This is would open up the opportunity to transition to `GetPot` for input file parsing down the road.  I haven't done that in this PR, though.

Tests are passing locally for me.

Feedback appreciated.

### DIFF
--- a/inc/queso/Makefile.am
+++ b/inc/queso/Makefile.am
@@ -27,6 +27,7 @@ BUILT_SOURCES += VectorSequence.h
 BUILT_SOURCES += VectorSet.h
 BUILT_SOURCES += VectorSpace.h
 BUILT_SOURCES += VectorSubset.h
+BUILT_SOURCES += BaseInputOptions.h
 BUILT_SOURCES += BasicPdfsBase.h
 BUILT_SOURCES += BasicPdfsBoost.h
 BUILT_SOURCES += BasicPdfsGsl.h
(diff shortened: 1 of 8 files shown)


dmcdougall:
@pbauman Does this look reasonable?  If so, I'll start adding documentation to the base class and work on some of the other options classes.

pbauman:
Good first step. :+1:

dmcdougall:
@pbauman Thanks.

PR #349, opened 2015-04-27
Default to silent rules for build system

pbauman:
Instead of having the full compile line emitted by default when
building, we will instead get something like
CXX target_name
Then, if we want a verbose build, you do make V=1 instead. This looks
better, IMHO, when building, but it also makes warnings stick out
like a sore thumb, which is also useful.


dmcdougall:
@pbauman Thanks.

PR #350, opened 2015-04-27
Refactor input options for Environment

dmcdougall:
Same deal as #347 and addresses part of #346.

There was a setter method `resetIdentifyingString` that changed the internal state of `Environment` but was labelled as `const`.  This has obviously been removed because having a `const` setter is totally unacceptable.  Furthermore, because this setter changed the internal state of `Environemnt` via the internally held `EnvOptionsValues` object, I've had to remove the promise that the input options object passed in the constructor is not changed.

Feedback welcome.


dmcdougall:
This looks good now.  If I don't have feedback from anyone else tomorrow and Travis comes back happy I'll merge this.

pbauman:
Nothing sticks out. :+1: Once I start exercising this, I'll scream if anything doesn't seem to be working.

PR #351, opened 2015-04-28
Interpolation Surrogate

pbauman:
Adds linear Lagrange interpolant surrogate for arbitrary dimension. Note I've defined a hierarchy of classes to promote an interface and reusability. In particular, I intended to create subclasses of the Gaussian likelihoods, for example, that take a pointer to `SurrogateBase`, then it becomes that much easier for the user to use surrogate models in their likelihood. I'd appreciate feedback on this point, particularly since there's some GP work underway already --- we'd want GP model surrogates to subclass this base. Future spectral projection surrogates would also subclass `SurrogateBase`.

@dmcdougall I'd appreciate you taking a good look through the comments and make sure that they make sense and you can follow the code. The equations to code translation may not be obvious everywhere here and I want to make sure the documentation is clear.

I still plan on adding a `Generator` (other name suggestions welcome) that will let the user supply their model and then QUESO can farm off filling in the values over subenvironments (in the QUESO parlance) over many processors. Also want to add constructors that take objects that will populate the surrogate. But at least this can start getting a review.

I have unit tests out to 4D. This won't be of much utility beyond that number of dimensions I don't think, but at least gives some warm and fuzzies about the correctness.

### DIFF
--- a/src/surrogates/inc/InterpolationSurrogateBase.h
+++ b/src/surrogates/inc/InterpolationSurrogateBase.h
@@ -60,10 +60,24 @@ namespace QUESO
         This function will return the corresponding global index to which the value
         at i,j,k can be indexed. Ordering must be consistent between coord_indices
         and n_points. Must be ordered in increasing dimension. e.g. x,y,z,...
-        The user shouldn't need to call this method, this is public mainly to facilitate
-        testing. */
+        The user shouldn't need to call this method, this is public mainly to
+        facilitate testing. */
     unsigned int coordToGlobal( const std::vector<unsigned int>& coord_indices,
                                 const std::vector<unsigned int>& n_points ) const;
+
+    unsigned int dim() const
+    { return this->m_domain.vectorSpace().dimGlobal(); };
+
+    //! Lower bound of domain along dimension dim
+    double x_min( unsigned int dim ) const
+    { return this->m_domain.minValues()[dim]; };
+
+    //! Upper bound of domain along dimension dim
+    double x_max( unsigned int dim ) const
+    { return this->m_domain.maxValues()[dim]; };
+
+    //! Spacing between points along dimension dim
+    double spacing( unsigned int dim ) const;
 
   protected:
 
--- a/src/surrogates/inc/InterpolationSurrogateBase.h
+++ b/src/surrogates/inc/InterpolationSurrogateBase.h
@@ -54,16 +54,6 @@ namespace QUESO
 
     virtual ~InterpolationSurrogateBase(){};
 
-    //! Map coordinate indices to a singal global index.
-    /*! e.g. in 3-D, you pass in i,j,k and the n_points in each of the 3 directions.
-        This function will return the corresponding global index to which the value
-        at i,j,k can be indexed. Ordering must be consistent between coord_indices
-        and n_points. Must be ordered in increasing dimension. e.g. x,y,z,...
-        The user shouldn't need to call this method, this is public mainly to
-        facilitate testing. */
-    unsigned int coordToGlobal( const std::vector<unsigned int>& coord_indices,
-                                const std::vector<unsigned int>& n_points ) const;
-
   protected:
 
     const InterpolationSurrogateData<V,M>& m_data;
--- a/src/surrogates/inc/InterpolationSurrogateBase.h
+++ b/src/surrogates/inc/InterpolationSurrogateBase.h
@@ -28,6 +28,7 @@
 #include <queso/SurrogateBase.h>
 #include <queso/BoxSubset.h>
 #include <queso/VectorSpace.h>
+#include <queso/InterpolationSurrogateBuilder.h>
 
 namespace QUESO
 {
@@ -52,6 +53,11 @@ namespace QUESO
     InterpolationSurrogateBase(const BoxSubset<V,M> & domain,
                                const std::vector<unsigned int>& n_points,
                                const std::vector<double>& values);
+
+    //! Constructor
+    /*! We extract data from the SurrogateBuilder. Note that the builder
+        MUST live as long as this object since we only extract const&. */
+    InterpolationSurrogateBase( const InterpolationSurrogateBuilder<V,M>& builder );
 
     virtual ~InterpolationSurrogateBase(){};
 
--- a/src/surrogates/inc/InterpolationSurrogateBase.h
+++ b/src/surrogates/inc/InterpolationSurrogateBase.h
@@ -41,7 +41,7 @@ namespace QUESO
       index coordinates (i,j,k,...), where i runs from (0, n_points[0]-1),
       j run from (0,n_points[1]-1), etc. We use this indexing to build maps. */
   template<class V, class M>
-  class InterpolationSurrogateBase : public SurrogateBase<V,M>
+  class InterpolationSurrogateBase : public SurrogateBase<V>
   {
   public:
 
--- a/src/surrogates/inc/InterpolationSurrogateBase.h
+++ b/src/surrogates/inc/InterpolationSurrogateBase.h
@@ -52,9 +52,9 @@ namespace QUESO
   public:
 
     //! Constructor
-    /*! n_points should be a vector with dimension matching the dimension of
-        the parameter space and contain the number of points in each coordinate
-        direction for the interpolant. */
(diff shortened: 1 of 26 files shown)


pbauman:
Rebased on dev. (serial) Tests and example added. I'm just about ready. There's a bug in the parallel case that I'm chasing, but once that's resolved, this is good to go. Criticism welcome, particularly about documentation. I'm hoping I commented enough that this is self-documenting, but let me know if it's not.

pbauman:
OK, I believe this is good to go. Figured out the parallel bug and the examples works for me on 1-8 subenviroments.

dmcdougall:
You need a [small modification to this variable here](https://github.com/libqueso/queso/blob/dev/doxygen/queso.dox.in#L568) in `queso.dox` so that your documentation will get rendered automatically by the build system.

src/surrogates/inc/SurrogateBase.h
```diff
@@ -0,0 +1,52 @@
...
+  class SurrogateBase
+  {
+  public:
+
+    SurrogateBase(){};
+
+    virtual ~SurrogateBase(){};
+
+    //! Method to return value given the parameter vector
+    virtual double evaluate(const V & domainVector) const =0;
```
dmcdougall: This looks as though we're only supporting surrogates of models that map `\mathbb{R}^n \to \mathbb{R}`, is that correct?  I.e., we're not currently supporting `\mathbb{R}^n \to \mathbb{R}^m`

I'm not suggesting implementing the multi-dimensional output case in this PR (if at all), I'm merely asking for purposes pertaining to my understanding.

src/surrogates/inc/InterpolationSurrogateBase.h
```diff
@@ -0,0 +1,69 @@
...
+//
+//-----------------------------------------------------------------------el-
+
+#ifndef UQ_INTERPOLATION_SURROGATE_BASE_H
+#define UQ_INTERPOLATION_SURROGATE_BASE_H
+
+#include <queso/SurrogateBase.h>
+#include <queso/BoxSubset.h>
+#include <queso/VectorSpace.h>
+#include <queso/InterpolationSurrogateData.h>
```
dmcdougall: This is up to you, but it looks like the only header you need here is the base class header.  You can get away with just a forward declaration of `InterpolationSurrogateData`.

src/surrogates/inc/InterpolationSurrogateBase.h
```diff
@@ -0,0 +1,69 @@
...
+
+namespace QUESO
+{
+  //! Base class for interpolation-based surrogates
+  /*! This class is used for surrogoate approximations of a model using interpolation.
+      Subclasses will define behavior of interpolant, but common assumptions
+      are: 1. Bounded domain; future work may extend behavior to unbounded domains.
```
dmcdougall: This list doesn't render properly in doxygen.

src/surrogates/inc/InterpolationSurrogateBase.h
```diff
@@ -0,0 +1,69 @@
...
+
+#include <queso/SurrogateBase.h>
+#include <queso/BoxSubset.h>
+#include <queso/VectorSpace.h>
+#include <queso/InterpolationSurrogateData.h>
+
+namespace QUESO
+{
+  //! Base class for interpolation-based surrogates
+  /*! This class is used for surrogoate approximations of a model using interpolation.
```
dmcdougall: Perhaps clarify the vector spaces the model maps from/to.

dmcdougall:
@pbauman I've looked through the documentation and design and I think this looks pretty solid.  I've still to go through and give you some more fine-grained feedback and things I need clarification on.

Thanks for taking the time to put this together; first impressions are it'll be very useful for small dimensional problems.

src/surrogates/inc/SurrogateBase.h
pbauman: Correct. If the user has multiple outputs, I envision creating one of these for each output. I'll update the documentation accordingly.

src/surrogates/inc/InterpolationSurrogateBase.h
pbauman: Good call. Thanks for catching my slop.

src/surrogates/inc/InterpolationSurrogateBase.h
pbauman: :-/ OK, thanks, I'll take a look.

pbauman:
>I've still to go through and give you some more fine-grained feedback and things I need clarification on.

Thanks for taking a look. I'll push a commit addressing the comments you made.

> first impressions are it'll be very useful for small dimensional problems.

Yeah, I can't see this really being useful beyond 4, maybe 5 dimensions, but if your science has 3 or less parameters, or even just to play with the problem a bit, this helps get you started. I plan to use this on the nitridation problem ASAP.

Future work will build around `SurrogateBase`, e.g. spectral projection (like we discussed at SIAM). Would like GPs, but I'm still not clear exactly what's going on there. I gather @roystgnr is working on GPMSA, but not sure of anything beyond that.

src/surrogates/inc/InterpolationSurrogateBase.h
pbauman: Hmm. I wonder if maybe the base class should be renamed to reflect that this is designed for single quantity output (vs. multiple quantity)?

pbauman:
Rebased and address @dmcdougall's current comments.

src/surrogates/inc/InterpolationSurrogateBase.h
```diff
@@ -0,0 +1,74 @@
...
+      For the structured grid, we think of referencing each "node" in the box by its
+      index coordinates (i,j,k,...), where i runs from (0, n_points[0]-1),
+      j run from (0,n_points[1]-1), etc. We use this indexing to build maps. */
+  template<class V, class M>
+  class InterpolationSurrogateBase : public SurrogateBase<V>
+  {
+  public:
+
+    //! Constructor
+    /*! n_points should be a vector with dimension matching the dimension of
```
dmcdougall: `data.n_points` instead of `n_points`?

src/surrogates/inc/LinearLagrangeInterpolationSurrogate.h
```diff
@@ -0,0 +1,110 @@
...
+  protected:
+
+    //! Helper function to get lower bound indices for each dimension
+    /*! domainVector contains the current value at which want want to interpolate.
+        We use that to figure out the bounding values in each dimension. Then
+        we populate indices with the lower index. */
+    void compute_interval_indices(const V & domainVector,
+                                  std::vector<unsigned int>& indices) const;
+
+    //! Helper function to populate bounding values for the intervales in each dimension
```
dmcdougall: Typo: intervales.

dmcdougall:
This is well documented, well tested and well written code.  Good job.

I've gone through the documentation and barring a few minor typos everything looks good.

I've gone through the code itself.  I'm not going to comment on or correct the code style because the code style throughout `queso` itself is not consistent.  I have some plans to amend in the future.  There are plenty of comments which make this a pleasure to read.

I've gone through the tests and I'm very happy to the extent this has been tested.  Well done.

I've also build the doxygen documentation and looked at how the classes are structured and what interface we're exposing to the user.  You've done a good job in making your interfaces extensible and leave everything open-ended for changing requirements and new surrogate functionality in the future.  At first I was going to suggest subclassing `BaseScalarFunction` for the `evaluate_model` method, but actually I think not doing that is more appropriate; I was confusing myself between *model* evaluations and *likelihood* evaluations.  In fact, you take a very similar approach to the one I took in coding up the canned Gaussian likelihood classes a while back.

This is a very big thumbs-up from me.  Thank you so much for taking the time and effort to put this together.

@pbauman Fix up the minor typos and don't forget to make a note in the changelog.  After you've done that I will merge this.

P.S.  Thank you for using the `queso_assert*` stuff.  Code is much more readable than using the old `UQ_FATAL_MACRO*` macros.

pbauman:
Fixed. Thanks for the review. I need to add a reader, writer, and checkpointing, but I'll do that in a separate PR.

pbauman:
Rebased

PR #352, opened 2015-04-28
Refactor input options for ExperimentModel

dmcdougall:
Addresses part of #346.

### DIFF
--- a/src/gp/inc/ExperimentModel.h
+++ b/src/gp/inc/ExperimentModel.h
@@ -62,18 +62,18 @@ public:
 
 private:
   // Private variables
-  const BaseEnvironment&           m_env;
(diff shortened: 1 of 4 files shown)


pbauman:
:+1: 

I just noticed that `BaseEnvironment::allOptonsMap ` - I'd like to see that moved into an intermediary-boost-options parser class, like we discussed some time ago. Just making a note here while I'm thinking about, doesn't affect merging this.

dmcdougall:
> I just noticed that `BaseEnvironment::allOptonsMap` - I'd like to see that moved into an intermediary-boost-options parser class, like we discussed some time ago. Just making a note here while I'm thinking about, doesn't affect merging this.

I agree with you.  I'll open a separate ticket for that.

Thanks for giving this a look over.

PR #354, opened 2015-04-29
GP fixups

roystgnr:
A couple minor tweaks to make the GP example easier to run


dmcdougall:
@roystgnr Thanks a lot.  I should really get in the habit of running the examples from the install directory to catch things like this.

PR #356, opened 2015-04-29
Refactor input options for GpmsaComputerModel

dmcdougall:
Addresses part of #346.


dmcdougall:
Rebased on the new asserts refactor.  This looks fine to me.  If nobody else has feedback here I'll merge when Travis comes back happy.

PR #357, opened 2015-04-29
Refactor input options for GPMSA

dmcdougall:
Addresses part of #346.


src/gp/inc/GPMSAOptions.h
```diff
@@ -45,9 +46,6 @@ class GPMSAOptions
   //! Destructor
  ~GPMSAOptions();
```
dmcdougall: I should have made this `virtual` for readability.  It's not necessary, but I'll push that change directly to `dev`.

dmcdougall: Change is [here](6c523a39ff9ae69d1f746790146e8e78357af8d5).

PR #359, opened 2015-04-30
Refactor input options for MLSamplingOptions

dmcdougall:
Addresses part of #346.

Note:  It doesn't look like this options class is used by anything.


dmcdougall:
This looks good to me.  If there's no feedback and Travis comes back happy I'll merge this tomorrow.

PR #364, opened 2015-05-01
Refactor input options for SequenceStatisticalOptions

dmcdougall:
Addresses part of #346.

### DIFF
--- a/src/basic/inc/SequenceStatisticalOptions.h
+++ b/src/basic/inc/SequenceStatisticalOptions.h
@@ -30,6 +30,7 @@
 #ifdef QUESO_USES_SEQUENCE_STATISTICAL_OPTIONS
 
 #include <queso/Environment.h>
+#include <queso/BaseInputOptions.h>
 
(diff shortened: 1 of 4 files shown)


dmcdougall:
Most of the statistical functionality is disabled throughout the codebase, but a lot of the deprecation and rebasing work is done here and looks fine to me.

If there's no feedback tomorrow and Travis comes back happy I'll merge this.

csim314:
Do we have plans to re enable  stats? Having to post process the same quantities repeatedly seems kinda lame. No?

pbauman:
> Having to post process the same quantities repeatedly seems kinda lame. No?

IMHO, we leave whatever functioning stats are there (are they tested?), but document that we have no plans to further develop "raw" statistics. Producing the MC chain is the hard part and I feel like we'd just be wasting developer time implementing new statistics when we could just export the chain (which we already do) and use Octave/SciPy/R, which will always have far more statistics sauce. A canned Matplotlib/SciPy script would be very useful in share to distribute. Again, all IMHO.

PR #368, opened 2015-05-03
Add ScopedPtr

pbauman:
In my interp-surrogate-writer branch, I want to be able to use a smart pointer to manage memory of pointers inside an object. `std::unique_ptr` is the natural choice, but that requires C++11. `boost::scoped_ptr` is there, but I'm hoping we can remove the Boost requirement one of these days. `std::auto_ptr` is deprecated. So, what this PR does is use a template typedef idiom to define ScopedPtr using `std::unique_ptr`, then `boost::scoped_ptr` if `std::unique_ptr` is not available, then `std::auto_ptr` if the previous two are not available. ([Similar idea for SharedPtr](http://stackoverflow.com/questions/2937351/how-to-use-autoconf-with-c0x-features?answertab=active#tab-top).) This covers the case when we enable C++11 (even optionally, `std::unique_ptr` has been around for awhile in the compilers) and the case if we don't require C++11, but remove the Boost mandatory dependency. I used the name `ScopedPtr` to indicate that a possible fallback is `boost::scoped_ptr` and that does not have move semantics (where as `std::unique_ptr` does). But once we require C++11, we can search/replace `ScopedPtr` to `std::unique_ptr` and remove this wrapper. I gather `std::unique_ptr` and `boost::unique_ptr` have some differences, but I haven't looked deeply into it, so I was just avoiding that altogether since I don't need move semantics, just a very basic smart pointer.

We do not yet test for C++11 and I've not added it, nor the test for `std::auto_ptr`; those autoconf tests are a PITA to write and I'm trying to move fast. But, we require boost, so this will work as is while Boost is mandatory. And the Boost autoconf macro had already wrapped up tests for headers, so there's now an autoconf check for `boost::scoped_ptr`.

Although there's no test, and this doesn't actually compile anything yet that uses it, I can vouch that I'm already using this on the aforementioned branch and objects that are using it are compiling. I wasn't sure if this would be controversial or not, so I gave it its own PR instead of tucking it inside the aforementioned branch.

An example of declaring this would look like `QUESO::ScopedPtr<GslVector>::Type vector;`

Thoughts @dmcdougall @roystgnr?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -123,6 +123,7 @@ AC_CACHE_SAVE
 AC_LANG([C++])
 BOOST_REQUIRE([1.35])
 BOOST_PROGRAM_OPTIONS
+BOOST_FIND_HEADER([boost/scoped_ptr.hpp])
 AC_CACHE_SAVE
 
 # Check for GLPK (optional)
--- a/inc/queso/Makefile.am
+++ b/inc/queso/Makefile.am
@@ -58,6 +58,7 @@ BUILT_SOURCES += OptimizerMonitor.h
 BUILT_SOURCES += RngBase.h
 BUILT_SOURCES += RngBoost.h
 BUILT_SOURCES += RngGsl.h
+BUILT_SOURCES += ScopedPtr.h
 BUILT_SOURCES += TeuchosMatrix.h
 BUILT_SOURCES += TeuchosVector.h
 BUILT_SOURCES += Vector.h
@@ -295,6 +296,8 @@ RngBase.h: $(top_srcdir)/src/core/inc/RngBase.h
 RngBoost.h: $(top_srcdir)/src/core/inc/RngBoost.h
 	$(AM_V_GEN)rm -f $@ && $(LN_S) $< $@
 RngGsl.h: $(top_srcdir)/src/core/inc/RngGsl.h
+	$(AM_V_GEN)rm -f $@ && $(LN_S) $< $@
+ScopedPtr.h: $(top_srcdir)/src/core/inc/ScopedPtr.h
 	$(AM_V_GEN)rm -f $@ && $(LN_S) $< $@
 TeuchosMatrix.h: $(top_srcdir)/src/core/inc/TeuchosMatrix.h
 	$(AM_V_GEN)rm -f $@ && $(LN_S) $< $@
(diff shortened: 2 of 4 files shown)


pbauman:
#369 has an example of using this.

dmcdougall:
>I can vouch that I'm already using this on the aforementioned branch and objects that are using it are compiling

Which `#ifdef` branch does this hit?  I.e., are you using the `boost` smart pointer or the one in the `std` namespace?  Have you tried both?

pbauman:
> Which #ifdef branch does this hit? I.e., are you using the boost smart pointer or the one in the std namespace? Have you tried both?

Sorry, I tried to explain this, but failed. We will hit Boost, until we start (or the user) adds `-std=c++11` to the compile line, then it will hit the first `#ifdef`. But the is that it's supposed to be invisible to the user.

dmcdougall:
> Sorry, I tried to explain this, but failed. We will hit Boost, until we start (or the user) adds -std=c++11 to the compile line, then it will hit the first #ifdef. But the is that it's supposed to be invisible to the user

On the contrary, your initial explanation was entirely clear.  It was my question that was not clear.  I'll try to rephrase.  You mentioned that you use this code in another one of your branches.  In that branch, what is `QUESO::ScopedPtr` after the C preprocessor has finished executing?

pbauman:
> In that branch, what is QUESO::ScopedPtr after the C preprocessor has finished executing?

Ah, OK, thanks for the clarification. It is  `boost::scoped_ptr`. This will always happen in the case that one does not have `-std=c++11` in their `CXXFLAGS` because QUESO currently requires Boost. Once we do not require Boost (whenever that is...), we'll need to add an m4 to check for `std::auto_ptr` so we can fall back to that until we require C++11, or someone gets around to adding a `std::unique_ptr` test.

dmcdougall:
Ok, and the last part of my question was, "Have you tried both?"  That is, did you try passing `-std=c++11` as part of your `CXXFLAGS` at configure-time to check that your other branch compiles when `QUESO::ScopedPtr` is `std::unique_ptr`?

pbauman:
Oh, sorry. I have passed `-std=c++11`, but that doesn't do anything at this point (as far as ScopedPtr goes), because we need to introduce a test to define the macro `QUESO_HAVE_CXX11_UNIQUE_PTR` and I haven't done that yet (we can change the macro name, I just made it up). I did at least look for something in the autoconf archive, but didn't see any. So we'll need to copy from someone else or write our own. But again, since we have a Boost requirement, we don't need it yet.

dmcdougall:
>  we need to introduce a test to define the macro `QUESO_HAVE_CXX11_UNIQUE_PTR` and I haven't done that yet

Ok cool.  So we can do that whenever we need it.

pbauman:
> because we need to introduce a test to define the macro QUESO_HAVE_CXX11_UNIQUE_PTR

One thing I'll point out for future reference. The m4 test we write should, itself, define `HAVE_CXX11_UNIQUE_PTR` (or whatever) --- the `QUESO_` part automatically gets added later via the `AX_PREFIX_CONFIG_H` macro configure.ac.

PR #369, opened 2015-05-03
InterpolationSurrogate I/O

pbauman:
Implements I/O classes for the `InterpolationSurrogateData`. Went ahead and made it standalone since it made it a little cleaner to be able to read and construct everything starting from a datafile and a `FullEnvironment`.

Note this is built on #368, once that's merged, I'll rebase this. If there are changes needed there, then I will update this PR accordingly.


examples/interpolation_surrogate/4d_interp.C
```diff
@@ -104,6 +105,12 @@ int main(int argc, char ** argv)
   // desired points in parameter space.
   builder.build_values();
 
+  // Now that we've built the data, we write it out so we can reuse it later
+  QUESO::InterpolationSurrogateIOASCII<QUESO::GslVector, QUESO::GslMatrix>
+    data_writer;
+
+  data_writer.write( "4d_interp_data.dat", data );
```
dmcdougall: What happens if you try to write/read something that hasn't yet been built by the builder?

src/surrogates/inc/InterpolationSurrogateData.h
```diff
@@ -83,6 +83,13 @@ namespace QUESO
     //! Get spatial coordinate value at the node index along coordinate direction dim
     double get_x( unsigned int dim, unsigned int index ) const;
 
+    //! Sync values across all processors from root processor
+    /*! m_values may only be set on one processor, to this method
```
dmcdougall: Typo: 'to this' should be 'so this'.

examples/interpolation_surrogate/4d_interp.C
pbauman: That's a good question. One would desire an error, but write now it won't, it'll happily write no values out. I'll fix that, thanks.

src/misc/src/StreamUtilities.C
```diff
@@ -0,0 +1,55 @@
...
+
+    in.get(c);
+    // std::isspace used to be std::isblank, but the latter requires C++11
+    // There are slight differences, but for our purposes, this should
+    // be OK.
+    while(std::isspace(c))in.get(c);
+
+    in.putback(c);
+
+    while (in.get(c), c==comment_start)
```
dmcdougall: It took me a few minutes to realise that while `get(char & c)` does not return a `bool` (it returns a `istream &`), the result is [cast to a `bool` and returns the opposite of `fail`](http://www.cplusplus.com/reference/ios/ios/operator_bool/).

src/surrogates/inc/InterpolationSurrogateData.h
pbauman: Will fix, thanks.

src/misc/src/StreamUtilities.C
```diff
@@ -0,0 +1,55 @@
...
+       in.getline (line, 255);
+
+    // put back first character of
+    // first non-comment line
+    in.putback (c);
```
dmcdougall: Perhaps this is super defensive programming, but can you check the stream's `badbit` after this operation?  See here: http://www.cplusplus.com/reference/istream/istream/putback/

src/misc/src/StreamUtilities.C
pbauman: Honestly, I just copied and pasted this (I think Ben originally wrote it). Do you want to me to make it more explicit or add a comment?

src/misc/inc/StreamUtilities.h
```diff
@@ -0,0 +1,44 @@
...
+namespace QUESO
+{
+  class StreamUtilities
+  {
+  public:
+
+    StreamUtilities(){};
+    ~StreamUtilities(){};
+
+    static void skip_comment_lines( std::istream &in, const char comment_start);
```
dmcdougall: Ah ok, this function doesn't skip *all* comment lines, but stops at the first line that doesn't begin with a comment character.  Would you mind adding a little doxygen comment here?

src/surrogates/src/InterpolationSurrogateIOASCII.C
```diff
@@ -0,0 +1,241 @@
...
+        output.open( filename.c_str() );
+
+        // Write simpler header comments
+        std::string header = "# Data for interpolation surrogate\n";
+        header += "# Format is as follows:\n";
+        header += "# dimension (unsigned int)\n";
+        header += "# n_points in each dimension\n";
+        header += "# x_min, x_max pairs for each dimension\n";
+        header += "# values for each point in parameter space\n";
+        header += "# values musted ordered in structured format.\n";
```
dmcdougall: "musted"?

src/misc/src/StreamUtilities.C
dmcdougall: Not necessarily, I was commenting so anyone else reading this didn't have to do the digging I did.

dmcdougall:
Once the `ScopedPtr` stuff is merged I think this is good to go.

Thanks @pbauman.

pbauman:
Rebased and I believe I addressed all your comments. Let me know if you see anything else.

Issue #358, opened 2015-04-30
Move `BaseEnvironment::allOptionsMap` into an intermediary-boost-options parser class


dmcdougall:

dmcdougall:
First mentioned [here](https://github.com/libqueso/queso/pull/352#issuecomment-97785604).

dmcdougall:
The `allOptionsMap` is protected in `BaseEnvironment`.  I'm okay with removing this and absorbing it into an intermediary Boost options parser class because I don't expect the user to subclass `BaseEnvironment` to create their own custom queso environment.

Sound reasonable?

pbauman:
:+1: 

I'm still puzzled why there's even a base class to begin with. There's a "Full" version and an "Empty" version. Who would ever use an "Empty" environment?

dmcdougall:
I don't know.  Perhaps we can deprecate the subclasses and rename `BaseEnvironment` to `FullEnvironment`.

pbauman:
I'd leave it as is. Who knows what the future holds.

dmcdougall:
#376 is merged, so I'm closing this.

PR #376, opened 2015-05-08
Isolate boost-specific optionsMap to parser

dmcdougall:
Addresses #358.

This isn't quite ready yet.  The boost-specific methods in the environment class need to be deprecated, and I need to add some documentation to the interface.


dmcdougall:
I think this is good to go now.

Note that right now the environment class still contains boost-specific members and parsing logic to maintain backwards compatibility.  I have, however, deprecated those methods.

Feedback appreciated.

pbauman:
Is there a way you could do composition instead of inheritance at the final step? E.g. not have `EnvironmentOptions` subclass the Boost parsing, but rather *use* it. So `EnvironmentOptions` contains the values that will be supplied, but doesn't require the user to use Boost? Otherwise, it's not possible for me to use a different parser, right? Or is this just a step in the bigger picture?

pbauman:
Or, maybe the other way around - have a `EnvironmentOptionsBase` (or something) subclass the `BaseOptions` and contain the values. This would then the be thing that's passed to the `Environment` constructor. Then, you have a `BoostEnvironmenOptions` which then uses Boost.

Backward compatibility may be an issue though.

Is it possible to encapsulate program options within a few function calls and then just use composition as in my previous comment?

roystgnr:
"doesn't require the user to use Boost" would seem to be nearly impossible while maintaining compatibility with old input files, but...

How complicated are the input files that boost::po understands and the fields which QUESO currently reads?  What I've looked at so far seems to be simple "foo = bar", i.e. a very strict subset of GetPot format.  I would personally love to get rid of the boost::po dependency and thereby enable easy compilation with GLIBCXX_DEBUG options turned on.  Right now that's only an option if you've got a hand-dbg-compiled boost installation to link against.

pbauman:
> "doesn't require the user to use Boost" would seem to be nearly impossible while maintaining compatibility with old input files

This is much less of an issue in QUESO because you can pass *Values classes to the corresponding objects with the values already set. We just never have because if you pass a NULL pointer to that object, boost::po is used behind the scenes to parse a input file formatted for boost::po.

So my dream is to keep that functionality possible for QUESO backward compatibility, but optional. I.e. On my side, I start manually constructing the *Values objects myself (using GetPot to parse only 1 combined GRINS/QUESO input file) and pass the object to QUESO so that Boost is never needed.

Imagine wrapping up the smart pointers in GRINS so that we don't need Boost. Then, if we can make Boost optional in QUESO, I DON'T HAVE TO COMPILE BOOST ON SUPERCOMPUTERS ANYMORE!

dmcdougall:
This approach (inheritance) ties down each of the `*Values` classes to be parsed by `boost`.  If we want to be in a position to transition to an optional `boost` dependency then I think a more appropriate solution is composition, as @pbauman mentioned.  Each `*Values` class would take an argument of type `BaseInputOptionsParser &` in a constructor which would conform to a general enough interface to allow the parsing.  By default, when the user doesn't want to set options programmatically `QUESO` would instantiate the `*Values` object with a `BoostInputOptionsParser &` in its constructor.  `BoostInputOptionsParser` would subclass `BaseInputOptionsParser` and implement the interface for the parsing.  This way, to transition to an optional `boost` dependency all we'd have to do is implement a `GetPotInputOptionsParser` that subclasses `BaseInputOptionsParser` and switch the default passed parser when `QUESO` needs to construct a `*Values` object.

Thoughts?

pbauman:
> This way, to transition to an optional boost dependency all we'd have to do is implement a GetPotInputOptionsParser that subclasses BaseInputOptionsParser and switch the default passed parser when QUESO needs to construct a *Values object.

I'd argue this isn't even necessary. Even if Boost isn't there, the user can construct and set their own `*Values` class and provide them to the appropriate QUESO object constructors. Indeed, this is what I plan on doing. I don't want to dissuade phasing in a GetPot parser interface if you want to go that route, but given a choice, I'd rather focus on locking down the Boost/QUESO interface.

dmcdougall:
I've thought about this some more, and I think composition, rather than inheritance, really is the right approach here.  My reasoning for this is that the `*Values` classes are not parsers, and so the interpretation of inheritance as an "is-a" relationship is not being satisfied here.  Instead a `*Values` class should *use* a parser, hence composition.

I'll update this pull request to reflect that pattern.

> I'd rather focus on locking down the Boost/QUESO interface.

Yeah I will do that first since we already require `boost`.

dmcdougall:
Alright, I think the new design is better.  Rather than inheriting from a parser, the *Values classes will *use* a parser object.

I'll add some tests for the most heavily used options objects just to make sure everything still gets set properly but initial feedback would be good if you think there's any red flags.  I also need to update the docs too, but at least all the current tests are passing now.

dmcdougall:
Ok, I think this is finally done, but it's pretty big.  The main point is that now *OptionsValues objects *use* a parser rather than inheriting from one.  All the old options classes have been deprecated.  All the new options classes use the neat common parsing logic code rather than reinventing the wheel every time you want to parse, say, and `unsigned int`.  I've also added more strict sanity checks on the input options that were previously only warnings.  A good example is `m_filteredChainLag`, which only printed a warning if it was less than 2 and then set it to 2 anyway.  I've replaced this with a `queso_require_greater_equal_msg` with a helpful message informing the user that it *must* be at least two.  There were other options with similar behaviour and these have also been updated.

The old GPMSA code hasn't been updated.  It isn't tested and it will be deprecated with the newer functionality that uses the underlying `StatisticalInverseProblem` machinery that the old didn't use.  In lieu of this I saved myself some time and didn't bother refactoring those classes.

I understand that asking for feedback on such a big patch is daunting.  I'd be happy if someone checked one of the options classes and the parser class for feedback.

roystgnr:
"uses a parser rather than inheriting from one" is good.  But there's actually a *loss* of modularity when switching includes from BaseInputOptions.h to BoostInputOptionsParser.h.  I guess getOption is templated so there's no way to define it as virtual in the base class?

roystgnr:
Other than that caveat the diffs look good, and I don't see any way to fix that concern so I'd say merge as-is.

pbauman:
This was more or less my major complaint. Is there no way that we can pass in a pointer to the input parser so that everything Boost is external? Or is this a step in that direction?

pbauman:
@dmcdougall assuaged me in an offline conversation. Merge at will.

dmcdougall:
That felt good.

dmcdougall:
This broke the case where the user would pass defaults.  The options object would get `delete`d twice.  I have a fix (and test) incoming soon.

dmcdougall:
The pull request in question is #383.

PR #377, opened 2015-05-11
Generalize Interpolation Surrogate

pbauman:
If the user is, say, solving a PDE and extracting multiple quantities from that PDE solution from which they will build multiple surrogates, it would be nice if the surrogate building infrastructure allowed that. So, I redid the API on the SurrogateBuilders to support this. Namely, I introduced a simple container object `InterpolationSurrogateDataSet` that creates consistent `InterpolationSurrogateData` objects. The SurrogateBuilder objects now take the DataSet object. Futhermore, the `evaluate_model` function got changed to a `void` function and now also takes an `std::vector` which the user populates; these are the values for each of the different functions the user is trying to interpolate. Everything else is unchanged. Once the builder is done, we can write/read each surrogate model individually, as before, and the user can use those as they please. Also updated the I/O to allow the user to supply the rank for reading/writing so that, if there's multiple processes available, the user can read/write each dataset on a different rank.

Tests and examples updated and passing. In particular, I manually tested the example on multiple MPI ranks and it is behaving as expected.

### DIFF
--- a/src/surrogates/inc/InterpolationSurrogateBuilder.h
+++ b/src/surrogates/inc/InterpolationSurrogateBuilder.h
@@ -61,12 +61,6 @@ namespace QUESO
     //! Cache the amount of work for each subenvironment
     std::vector<int> m_njobs;
 
-    //! Ensure that if fullComm() size > 1, then n_subenvironments > 1
-    /*! If not, this is a strange configuration and breaks things like inter0Comm
-        and would induce redundant work. Thus, this configuration is not supported and
-        we error out. */
-    void check_process_config();
-
     //! Partition the workload of model evaluations across the subenvironments
     void partition_work();
 
--- a/src/surrogates/inc/InterpolationSurrogateBuilder.h
+++ b/src/surrogates/inc/InterpolationSurrogateBuilder.h
@@ -26,7 +26,7 @@
 #define UQ_INTERPOLATION_SURROGATE_BUILDER_H
 
 #include <queso/SurrogateBuilderBase.h>
-#include <queso/InterpolationSurrogateData.h>
+#include <queso/InterpolationSurrogateDataSet.h>
 
 namespace QUESO
 {
@@ -47,7 +47,7 @@ namespace QUESO
     //! Constructor
     /*! We do not take a const& to the data because we want to compute and
         set the values directly. */
-    InterpolationSurrogateBuilder( InterpolationSurrogateData<V,M>& data );
+    InterpolationSurrogateBuilder( InterpolationSurrogateDataSet<V,M>& data );
 
     virtual ~InterpolationSurrogateBuilder(){};
 
@@ -56,7 +56,7 @@ namespace QUESO
 
   protected:
 
-    InterpolationSurrogateData<V,M>& m_data;
+    InterpolationSurrogateDataSet<V,M>& m_data;
 
     //! Cache the amount of work for each subenvironment
     std::vector<int> m_njobs;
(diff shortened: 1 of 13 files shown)


src/surrogates/inc/InterpolationSurrogateBuilder.h
```diff
@@ -47,7 +47,7 @@ namespace QUESO
     //! Constructor
     /*! We do not take a const& to the data because we want to compute and
         set the values directly. */
-    InterpolationSurrogateBuilder( InterpolationSurrogateData<V,M>& data );
+    InterpolationSurrogateBuilder( InterpolationSurrogateDataSet<V,M>& data );
```
dmcdougall: Why replace the original one instead of *adding* the functionality?  The difference being I wouldn't have to populate an `std::vector` with one element.  I guess that's not too much of a hassle...

What do you think?

src/surrogates/inc/InterpolationSurrogateDataSet.h
```diff
@@ -0,0 +1,89 @@
...
+
+    unsigned int size() const
+    { return m_datasets.size(); }
+
+  private:
+
+    InterpolationSurrogateDataSet();
+
+    //! Data structure to hold all data sets
+    std::vector<InterpolationSurrogateData<V,M>*> m_datasets;
```
dmcdougall: Could we use the new `ScopedPtr` infrastructure here instead of a raw pointer?  If so, should we?

src/surrogates/inc/SurrogateBuilderBase.h
```diff
@@ -41,7 +44,7 @@ namespace QUESO
 
     virtual ~SurrogateBuilderBase(){};
 
-    virtual double evaluate_model( const V & domainVector ) const =0;
+    virtual void evaluate_model( const V & domainVector, std::vector<double>& values ) =0;
```
dmcdougall: As I mentioned before, if I only have one data set I'd have to populate a vector of length 1, but maybe this isn't a big deal.

src/surrogates/src/InterpolationSurrogateDataSet.C
```diff
@@ -0,0 +1,55 @@
...
+    for( unsigned int s = 0; s < n_datasets; s++ )
+      m_datasets[s] = new InterpolationSurrogateData<V,M>( domain, n_points );
+  }
+
+  template<class V, class M>
+  InterpolationSurrogateDataSet<V,M>::~InterpolationSurrogateDataSet()
+  {
+    for( typename std::vector<InterpolationSurrogateData<V,M>*>::iterator it = m_datasets.begin();
+         it != m_datasets.end(); ++it )
+      delete *it;
```
dmcdougall: Just to reiterate, we wouldn't need to do this if the vector contained instances of `ScopedPtr` objects.

dmcdougall:
@pbauman I left some comments inline in the code.  They're not necessarily suggestions but more asking for clarification.  The tests pass and I'm happy for this to be merged, especially since you're already using it in production.

src/surrogates/inc/InterpolationSurrogateBuilder.h
pbauman: My thinking was that the single version would be the exception rather than the rule and that it's more code to maintain. But, if you would rather both be kept around, I'll go back and add.

src/surrogates/inc/InterpolationSurrogateDataSet.h
pbauman: Unfortunately, [you can't use scoped_ptr in a vector](http://stackoverflow.com/questions/17132938/undefined-reference-while-vector-of-boostscoped-ptr) since it doesn't have copy/move semantics. So, we'd need a `unique_ptr` to be able to use in an `std::vector`, but we don't have a wrapper for that (and it's more involved since we need something to fallback on when it's not available), so I just left it as a raw pointer. Perhaps I should add a `\todo` to flag it when when we move to requiring C++11?

src/surrogates/inc/InterpolationSurrogateBuilder.h
dmcdougall: No, I think what you have is fine.  Filling a vector with one element isn't a big deal and I imagine most use-cases involve several quantities to calibrate rather than just one.

src/surrogates/inc/InterpolationSurrogateDataSet.h
dmcdougall: Thanks for the clarification.

PR #379, opened 2015-05-20
Disable HDF5-dependent examples if HDF5 is not configured

roystgnr:
Long term, I need to upgrade our hdf5.m4 so that libMesh and QUESO (and anyone else) can use the Ubuntu 15.04 install.

Short term, this at least gets us compiling without errors again.


roystgnr:
Hold on, looks like I missed one

roystgnr:
Okay, now "make check" works for me again.

dmcdougall:
Thanks.

PR #380, opened 2015-05-26
Don't forget MPI_Finalize() if !QUESO_HAVE_HDF5

roystgnr:
Otherwise MPI returns 1 and run_all.sh thinks we have an actual error
rather than just a disabled example.


RhysU:
atexit that right after MPI init.

- Rhys

roystgnr:
atexit requires a function returning void, no?  MPI_Finalize returns int.  So I'd need a shim function, which would be just as much hassle as the duplicate call.

PR #381, opened 2015-05-29
Document all input file options for DRAM

dmcdougall:
This pull request has been a long time coming.  It documents all the input file options for DRAM; each option's default value; and any possible interaction with other input options where the interaction is not immediately obvious.

Doing this was painful as it required looking how each option played a role in the ~800 line `for` loop that is our DRAM algorithm, but this should help our users greatly.  Any feedback is graciously welcome.

This work branches from the refactor in #376.  PR #376 must be merged before this one.

### DIFF
--- a/src/stats/inc/MetropolisHastingsSGOptions.h
+++ b/src/stats/inc/MetropolisHastingsSGOptions.h
@@ -562,68 +562,118 @@ public:
 private:
   BoostInputOptionsParser * m_parser;
 
+  //! Option name for MhOptionsValues::m_help.  Option name is m_prefix + "mh_help"
   std::string                   m_option_help;
 
+  //! Option name for MhOptionsValues::m_dataOutputFileName.  Option name is m_prefix + "mh_dataOutputFileName"
   std::string                   m_option_dataOutputFileName;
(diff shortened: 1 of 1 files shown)


dmcdougall:
I should probably have stated this earlier, but the *new* commits on top of #376 are 745af42, 3566ef8, and 7ae5869.

Rather than looking at the net diff (which includes all the commits from #376), looking at each of the 3 commits listed above individually is much less laborious.

dmcdougall:
Pinging @briadam since you asked about input file options.  I'll follow up via email but I wanted to bring your attention to this for the sake of completeness.

dmcdougall:
Rebased since #376 was merged.

@nicholasmalaya Mind taking a look at this?

nicholasmalaya:
@dmcdougall I'll check it out tonight, will update.

dmcdougall:
@nicholasmalaya Think I've addressed most of your feedback.  For the no-op options I think it's best to keep them for now and remove them after the next release.  If you'd rather I remove them before the release I think I'd at least prefer to do it in a separate pull request.

nicholasmalaya:
I don't think we have to remove them. I just wanted to note that they existed, and that its probably worthwhile for one of us to kill 'em at a later date.

nicholasmalaya:
Fire the torpedoes.

Issue #203, opened 2014-02-09
Error message about Cholesky factorization of covariance matrix


lpswiler:
We have been seeing an error message while running QUESO.  I seem to remember that Ernesto said this wasn't really an error, but it is disconcerting, especially since we are giving QUESO proposal covariances that are positive definite.

This is the error message: 
In GslMatrix::chol(): iRC = 1, gsl error message = input domain error 
UQ RC ERROR, rank 0, in GslMatrix::chol(): matrix is not positive definite, iRc = 1 In 
MetropolisHastingsSG<P_V,P_M>::generateFullChain(): first chol failed

We would like to ensure this is not a true error and fix it in some way to remove the error message.

Thank you, 
Laura Swiler (Sandia Labs)

briadam:
@dmcdougall, @lpswiler  I will try to generate a small example that doesn't require a Dakota driver.  This is again our friend the cantilever beam, with one parameter order unity and another 1e7.  

It looks to me like the initial proposal covariance factors fine, but then after 3 or 4 chain samples (depending on initial covariance), the Cholesky "fails" on the adapted proposal covariance.  It appears that there's recourse via regularization in MetropolicHastingsSG::generateFullChain(), around line 2200, though I don't see the diagonal multiplier ever get used. Perhaps the initial factorization is failing and then life is good after adding diagonal weight.

Perhaps "failed" is just too strong a word to use in the output?

briadam:
@dmcdougall I was able to create a simple version of the Rosenbrock problem that demonstrates this error.  I created this branch: https://github.com/briadam/queso/tree/chol_error_203, with one modified test file (SHA: 4d8bb057ea53835ad533c8981f496b6e08eecdbe).  The following steps:
```
cd build/test
make test_LlhdTargetOutput
./test_LlhdTargetOutput test_StatisticalInverseProblem/llhdout_input.txt
```
Will yield:
```
In GslMatrix::chol(): iRC = 1, gsl error message = input domain error
Here is the offending matrix: 
0.111429 -0.265435 ; -0.265435 nan 
UQ RC ERROR, rank 0, in GslMatrix::chol(): matrix is not positive definite, iRc = 1
In MetropolisHastingsSG<P_V,P_M>::generateFullChain(): first chol failed
```

If the adapted proposal should remain valid for these adaptation settings, perhaps there is a bug.  If not and the recourse is expected, could we change the "error" to a warning?

dmcdougall:
[...]

PR #382, opened 2015-06-02
Refactor adaptive Metropolis logic

dmcdougall:
The main point of this pull request is to refactor the adaptive Metropolis code.

1. The adaptive Metropolis logic sat buried in an ~800 line `for` loop.  It is now in a separate function.
2.  The conditions required before executing the adaptation were tested and most of the adaptation code (300 lines) was nested in this `if` statement.  Where possible, I have negated this conditions so they become **exit** conditions, and de-dented the rest of the code.  This makes it a little easier to read.
3.  I have improved some of the printed warning messages that were raised in #203 to explain a little more what QUESO is doing.

All tests pass.  Feedback welcome.

### DIFF
--- a/src/stats/inc/MetropolisHastingsSG.h
+++ b/src/stats/inc/MetropolisHastingsSG.h
@@ -235,6 +235,7 @@ private:
                                    ScalarSequence<double>*      workingLogLikelihoodValues,
                                    ScalarSequence<double>*      workingLogTargetValues);
 
+  //! Adaptive Metropolis method that deals with adapting the proposal covariance matrix
   void adapt(unsigned int positionId,
       BaseVectorSequence<P_V, P_M> & workingChain);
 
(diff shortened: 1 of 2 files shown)


dmcdougall:
@briadam This should make the printed messages a little clearer.  You can find them [here](https://github.com/libqueso/queso/pull/382/files#diff-9c2da1f9c40803cdd4c137adcafd2a89R2314) and [here](https://github.com/libqueso/queso/pull/382/files#diff-9c2da1f9c40803cdd4c137adcafd2a89R2379) if you want to see them before checking out this branch.

dmcdougall:
@nicholasmalaya I've updated to reflect your input.  Look good?

nicholasmalaya:
> @nicholasmalaya I've updated to reflect your input. Look good?

Absolutely, thanks for the efforts.

PR #383, opened 2015-06-03
Fix double `delete` seg fault for various options objects

dmcdougall:
For user-provided options objects, we were doing a `delete`.  If the user also `delete`s their options object outside then a segmentation fault will occur.  I've just added a flag to check for when the user doesn't pass in an options object and only do the delete if the flag is true.


dmcdougall:
This fixed an issue introduced by #376.

dmcdougall:
Alright, this isn't the cleanest solution, but it fixes the problem for the time being.  A better approach would be to not deal with the pointer at all and just copy the options object (since it isn't big and it's only done once).  What I'll do is I'll merge this so that `dev` isn't broken, and I'll open a ticket for options object pointer cleanup for the future so I don't forget.

PR #390, opened 2015-06-18
Warning fixes

roystgnr:
This fixes all the trivial warnings clang gives me, as well as one serious warning with an obvious fix.

There are two serious warnings with non-obvious fixes; I just upped the volume on those so we don't forget them just because we're not regularly using clang.


dmcdougall:
Looks good, thanks.

PR #391, opened 2015-06-23
GslVector optimizations

roystgnr:
This turns on GSL's accessor inlining, inlines our operator[] shim, and turns off (except when DEBUG is defined) GSL's range checking.  Speeds up my GPMSA runs by ~20%.

### DIFF
--- a/src/core/inc/Defines.h
+++ b/src/core/inc/Defines.h
@@ -50,6 +50,14 @@
 //! This define is deprecated.  Remove any #ifdef statements in user code.
 #define QUESO_EXPECTS_LN_LIKELIHOOD_INSTEAD_OF_MINUS_2_LN
 
+// Use GSL inline functions
+#define HAVE_INLINE
+
+// And only do GSL range-checking if we're really debugging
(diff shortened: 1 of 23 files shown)


pbauman:
:+1:

src/core/inc/Defines.h
```diff
@@ -50,6 +50,14 @@
 //! This define is deprecated.  Remove any #ifdef statements in user code.
 #define QUESO_EXPECTS_LN_LIKELIHOOD_INSTEAD_OF_MINUS_2_LN
 
+// Use GSL inline functions
+#define HAVE_INLINE
+
+// And only do GSL range-checking if we're really debugging
+#ifndef DEBUG
```
dmcdougall: Shouldn't this be `NDEBUG`?  Or at least `#ifdef DEBUG`.  Please look up which of these patterns we're using, because I am not entirely certain.

dmcdougall: Oh wait, sorry, I misunderstood.  I think this is fine.

PR #392, opened 2015-07-18
Add 'txt' option for `mh_rawChain_dataOutputFileType`

dmcdougall:
These changes allow the user to specify `ip_mh_rawChain_dataOutputFileType = txt` in the `QUESO` input file.

Upon doing so, samples are written as space-delimited tabular data, with the first line specifying the number of samples and the number of parameters (in that order).  This option essentially behaves the same way as matlab output, but all the matlab-specific things are stripped.

The use-case for this option is that users not using matlab have an option of reading in tabular data without having to parse out matlab-specific boilerplate.

The default is still 'm', so behaviour doesn't change underneath the user.


dmcdougall:
Any objections here?  If I don't hear anything by tomorrow I'll go ahead and merge this.

dmcdougall:
> by tomorrow

I meant by the *end* of tomorrow.

pbauman:
Gimme gimme gimme. :+1:

pbauman:
I guess the is M-H specific and doesn't apply to ML?

dmcdougall:
Oh good catch.  Adding the ML option shouldn't be hard.

dmcdougall:
Looks like I already supported ML, since all the `(sub|unified)write*` methods are ushered through `ScalarSequence` or `SequenceOfVectors`.  I've updated the docs to better reflect that, though.

pbauman:
:+1:

PR #393, opened 2015-08-15
Make MPI optional

dmcdougall:
MPI is now optional and can be disabled at configure-time with a `--disable-mpi` flag.  When setting up the `FullEnvironment`, you can pass `0` as the communicator.  All the examples and tests have been updated to use this pattern.

Tests are passing in serial (without MPI) and in parallel.

@briadam @mseldre Would you mind trying this branch out and letting me know if this works for you?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -211,3 +211,6 @@ test/output_test_optimizer_options
 test/test_optimizer/input_test_optimizer_options
 test/test_optimizer_options
 test/test_SharedPtr
+test/output_test_serialEnv
+test/test_Environment/input_test_serialEnv
+test/test_serialEnv
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,5 +7,9 @@ before_script:
   - sudo apt-get install -q build-essential
   - sudo apt-get install -q libgsl0-dev
   - sudo apt-get install -q openmpi-bin openmpi-dev
-  - sudo apt-get install -q libboost-all-dev
-script: ./bootstrap && CC="mpicc" CXX="mpicxx" ./configure && make -j2 && make check
+  - sudo apt-get install -q libboost-dev libboost-math-dev libboost-program-options-dev
+script:
+  - ./bootstrap
+  - ./configure CC="mpicc" CXX="mpicxx"
+  - make -j4
+  - make check -j4
--- a/CHANGES
+++ b/CHANGES
@@ -3,6 +3,9 @@ QUESO: Quantification of Uncertainty for Estimation,
 Simulation, and Optimization.
 -----------------------------------------------------
 
+Version 0.54.0
+  * Make MPI an optional dependency
+
 Version 0.53.0
   * Add linear interpolation surrogates
   * Refactor input options processing
--- a/configure.ac
+++ b/configure.ac
@@ -92,15 +92,20 @@ AC_PROG_CXX
 AC_PROG_FC
 AC_LANG([C])
 
+HAVE_MPI=0
 if test "x$enable_mpi" = "xyes"; then
     ACX_MPI(CC="$MPICC", [AC_MSG_ERROR([Could not find MPI.])])
+    HAVE_MPI=1
 fi
 
 AC_LANG([C++])
 
 if test "x$enable_mpi" = "xyes"; then
     ACX_MPI(CXX="$MPICXX", [AC_MSG_ERROR([Could not find MPI.])])
+    HAVE_MPI=1
 fi
+AC_SUBST(HAVE_MPI)
+AM_CONDITIONAL(MPI_ENABLED, test x$HAVE_MPI = x1)
 
 #-------------------------
 # External Library Checks
--- a/configure.ac
+++ b/configure.ac
@@ -235,6 +235,7 @@ AC_CONFIG_FILES([
   test/test_InputOptionsParser/test_options_bad.txt
   test/test_InputOptionsParser/test_options_default.txt
   test/test_optimizer/input_test_optimizer_options
+  test/test_Environment/input_test_serialEnv
 ])
 
 AC_CONFIG_FILES([test/test_Regression/test_cobra_samples_diff.sh
--- a/configure.ac
+++ b/configure.ac
@@ -75,17 +75,32 @@ else
    AC_MSG_RESULT(no)
 fi
 
+#############################################
+# Check if the user explicitly disabled mpi #
+#############################################
+AC_ARG_ENABLE(mpi,
+              AS_HELP_STRING([--disable-mpi],
+                             [build without message passing support]),
+              [enable_mpi=no],
+              [enable_mpi=yes])
+
 #-------------------
 # Compilers and MPI
 #-------------------
-
 AC_PROG_CC
 AC_PROG_CXX
 AC_PROG_FC
 AC_LANG([C])
-ACX_MPI([CC="$MPICC"], [AC_MSG_ERROR([Could not find MPI.])])
+
+if test "x$enable_mpi" = "xyes"; then
+    ACX_MPI(CC="$MPICC", [AC_MSG_ERROR([Could not find MPI.])])
+fi
+
 AC_LANG([C++])
-ACX_MPI([CXX="$MPICXX"], [AC_MSG_ERROR([Could not find MPI.])])
+
+if test "x$enable_mpi" = "xyes"; then
+    ACX_MPI(CXX="$MPICXX", [AC_MSG_ERROR([Could not find MPI.])])
+fi
 
 #-------------------------
 # External Library Checks
--- a/examples/bimodal/src/bimodal_main.C
+++ b/examples/bimodal/src/bimodal_main.C
@@ -27,14 +27,20 @@
 int main(int argc, char* argv[])
 {
   // Initialize environment
+#ifdef QUESO_HAS_MPI
   MPI_Init(&argc,&argv);
   QUESO::FullEnvironment* env = new QUESO::FullEnvironment(MPI_COMM_WORLD,argv[1],"",NULL);
+#else
+  QUESO::FullEnvironment* env = new QUESO::FullEnvironment(argv[1],"",NULL);
+#endif
 
   // Compute
   compute(*env);
 
   // Finalize environment
   delete env;
+#ifdef QUESO_HAS_MPI
   MPI_Finalize();
+#endif
   return 0;
 }
--- a/examples/gaussian_likelihoods/blockDiagonalCovariance.C
+++ b/examples/gaussian_likelihoods/blockDiagonalCovariance.C
@@ -60,9 +60,12 @@ public:
 };
 
 int main(int argc, char ** argv) {
+#ifdef QUESO_HAS_MPI
   MPI_Init(&argc, &argv);
-
   QUESO::FullEnvironment env(MPI_COMM_WORLD, argv[1], "", NULL);
+#else
+  QUESO::FullEnvironment env(argv[1], "", NULL);
+#endif
 
   QUESO::VectorSpace<QUESO::GslVector, QUESO::GslMatrix> paramSpace(env,
       "param_", 1, NULL);
@@ -121,7 +124,9 @@ int main(int argc, char ** argv) {
 
   ip.solveWithBayesMetropolisHastings(NULL, paramInitials, &proposalCovMatrix);
 
+#ifdef QUESO_HAS_MPI
   MPI_Finalize();
+#endif
 
   return 0;
 }
--- a/examples/gaussian_likelihoods/blockDiagonalCovarianceRandomCoefficients.C
+++ b/examples/gaussian_likelihoods/blockDiagonalCovarianceRandomCoefficients.C
@@ -60,9 +60,12 @@ public:
 };
 
 int main(int argc, char ** argv) {
+#ifdef QUESO_HAS_MPI
   MPI_Init(&argc, &argv);
-
   QUESO::FullEnvironment env(MPI_COMM_WORLD, argv[1], "", NULL);
+#else
+  QUESO::FullEnvironment env(argv[1], "", NULL);
+#endif
 
   // Need 3 dims because two are for the hyperparameters
   QUESO::VectorSpace<QUESO::GslVector, QUESO::GslMatrix> paramSpace(env,
@@ -127,7 +130,9 @@ int main(int argc, char ** argv) {
 
   ip.solveWithBayesMetropolisHastings(NULL, paramInitials, &proposalCovMatrix);
 
+#ifdef QUESO_HAS_MPI
   MPI_Finalize();
+#endif
 
   return 0;
 }
--- a/examples/gaussian_likelihoods/diagonalCovariance.C
+++ b/examples/gaussian_likelihoods/diagonalCovariance.C
@@ -59,9 +59,12 @@ public:
 };
 
 int main(int argc, char ** argv) {
+#ifdef QUESO_HAS_MPI
   MPI_Init(&argc, &argv);
-
   QUESO::FullEnvironment env(MPI_COMM_WORLD, argv[1], "", NULL);
+#else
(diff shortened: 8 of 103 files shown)


nicholasmalaya:
Nice! Very cool feature to have. Would you also update the changelog to note this?

briadam:
@dmcdougall This is wonderful; will look forward to trying it out!

One note: I recently went through some pain with our typedefs that replace MPI types when Dakota is compiled serial.  You might double check that 0 is convertible to type MPI_Comm for the implementations you care about.  I am pretty sure it's valid for MPICH, but not sure about OpenMPI, in which MPI_Comm is either a structure or pointer to structure.  

I guess this would only matter though for someone building QUESO _with_ MPI, but then wanting to run it in serial (as Dakota will attempt to do).  In that case, we'd have a valid MPI_COMM_NULL from mpi.h that could be passed.  We'll just need to test that path.

dmcdougall:
>You might double check that 0 is convertible to type MPI_Comm for the implementations you care about. I am pretty sure it's valid for MPICH, but not sure about OpenMPI, in which MPI_Comm is either a structure or pointer to structure.

This can't happen because [we typedef `MPI_Comm` to `int` when queso is built without MPI](https://github.com/libqueso/queso/pull/393/files#diff-bc9006557478b21bac1f3d9ba78c5d2cR57).

>I guess this would only matter though for someone building QUESO with MPI, but then wanting to run it in serial (as Dakota will attempt to do). In that case, we'd have a valid MPI_COMM_NULL from mpi.h that could be passed. We'll just need to test that path.

If queso is built with MPI, you are forced to pass a communicator to queso.  If queso just so happens to be run serially, meaning the world communicator is of size 1, then there is a valid communicator to pass to queso.

There's a code path I was missing that @nicholasmalaya helped me pin down.  If queso is built without MPI but there is a user application that *is* built with MPI then it's possible the user could try to pass a valid MPI communicator to queso.  In this case, if it's not `int`-castable you'll get a compile-time error.  If it *is* `int`-castable then that's the integer that represents the communicator and since, if queso is built without MPI, it isn't used internally anyway there's nothing to worry about.  That said, we could be more defensive here and actually `#ifdef` around the `FullEnvironment` constructor to remove the communicator argument in serial.  This would prevent any garbage being passed in serial.  What do you think about this?

In parallel, if the user passes a garbage communicator then I think the 'garbage in garbage out' principle is a safe one to follow.  Since checking the type of the communicator (faux or otherwise) that is passed is stack-dependent I don't think queso should be responsible for this logic.

Have I missed anything?

pbauman:
> That said, we could be more defensive here and actually #ifdef around the FullEnvironment constructor to remove the communicator argument in serial.

If it's as simple as this, I would prefer this.

> Since checking the type of the communicator (faux or otherwise) that is passed is stack-dependent I don't think queso should be responsible for this logic.

Agree 100% here.

mseldre:
From Dakota, we need to worry about the four combinations of built with/without MPI times run with/without mpiexec, where 3 of the 4 are useful cases (serial executable launched with mpiexec should also work, but isn't useful unless you like repetition...).  The case that Brian and I are concerned about is the built with MPI + launched in serial without mpiexec.  In this case, since QUESO is built with MPI, we will need to pass it an MPI_Comm and we won't have called MPI_Init since we didn't detect a parallel launch.  In this case, Brian states that MPI_COMM_NULL should be valid, but this could potentially cause errors downstream (e.g., MPI_Comm_size(MPI_COMM_NULL)?).  It seems preferable for Dakota to be able to call a serial QUESO constructor, even when compiled with MPI support.  Would that be awkward?

dmcdougall:
>The case that Brian and I are concerned about is the built with MPI + launched in serial without mpiexec. In this case, since QUESO is built with MPI, we will need to pass it an MPI_Comm and we won't have called MPI_Init since we didn't detect a parallel launch. In this case, Brian states that MPI_COMM_NULL should be valid, but this could potentially cause errors downstream (e.g., MPI_Comm_size(MPI_COMM_NULL)?).

If you want to build queso with MPI support but you don't want to call `MPI_Init`, then the correct (and defined) communicator to pass is `MPI_COMM_SELF` (thanks for the tip @roystgnr).  The `MPI_COMM_NULL` communicator is a handle that represents an invalid (but defined in the absence of `MPI_Init`) communicator.  `libmesh` does exactly this.  Their communicator *defaults* to `MPI_COMM_SELF` at a point where it has no idea whether or not `MPI_Init` was called.

All we'd have to do in queso (to avoid doing a possibly unsafe `inputComm == MPI_COMM_SELF`) is make the constructor I was talking about `#ifdef`ing available when queso was built with MPI, set the default communicator to MPI_COMM_SELF, set the default communicator size to 1, and set the default rank to 0.  The parallel MPI routines would then do a run-time check for a communicator size of greater than 1.

>It seems preferable for Dakota to be able to call a serial QUESO constructor, even when compiled with MPI support. Would that be awkward?

I'm working on it now.

mseldre:
This was our thinking as well -- just pass MPI_COMM_SELF when running an MPI executable in serial mode.  This is how Dakota/QUESO is configured currently.  As you saw in San Diego, though, this case was failing when MPI_COMM_SELF is passed without supporting MPI_Init.  If the standard says it should work, then we can consider digging further to see why it was failing (assuming the point is not rendered moot).  I just did a quick review of mpi.h, but both MPI_COMM_SELF and MPI_COMM_NULL are buried in several layers of macros.

dmcdougall:
>As you saw in San Diego, though, this case was failing when MPI_COMM_SELF is passed without supporting MPI_Init.

Because we don't currently check the communicator size.  This means that in parallel, operations like `Gather` are still executed.  Executing operations on the communicator when `MPI_Init` was not called is not allowed.  This will hopefully no longer fail once we do the run-time check that the size is greater than one.

roystgnr:
*Passing* MPI_COMM_SELF to a function should compile regardless of whether MPI_Init() gets called.  But *using* MPI_COMM_SELF in any MPI calls is not guaranteed to work, so the best we can do with it is a "if (my_comm == MPI_COMM_SELF) mpi_code(); else serial_code();" test.

And at that point we might as well use a dedicated serial initialization API, set a saner run_without_mpi boolean for that test, and not have to worry about whether MPI_COMM_SELF == MPI_COMM_SELF is guaranteed to work or not.

We did this almost by accident in libMesh.  For optimization reasons our Parallel::Communicator class caches the size and rank information from the communicator, and avoids redundant MPI calls, and in the MPI_Init-never-called cases it naturally turned out that all the MPI calls were redundant.

mseldre:
Agreed.  Either an overloaded constructor (Comm or no Comm) or a single constructor with default Comm arguments seem like clean solutions for the serialized API.

dmcdougall:
Alright, I think this is in good shape.  I've added a (very) simple test that we can call the constructor in parallel and run an inverse problem.  We're not testing the output; just that no errors are thrown during the test.

@mseldre @briadam Feedback appreciated if you have some time to exercise this branch before merging.

pbauman:
Just a thought: do we want to squash f4c7f5c and 7c4cdfe so `make check` is clean in the history?

dmcdougall:
Done.

briadam:
You rock!  Nice work.  I will try this out after returning from travel.

briadam:
Initial testing with Dakota looks encouraging.  The following cases work (Dakota can't currently pass QUESO a comm of size > 1):
 * Compile Dakota/QUESO with MPI, mpirun -n 1
 * Compile Dakota/QUESO without MPI (serial run)

However this one fails:
 * Compile Dakota/QUESO with MPI, serial run

In this case, Dakota/QUESO are compiled with MPI, but then I construct the QUESO::FullEnvironment using the serial constructor (no MPI comm).  I get the following error after 1632 likelihood evals:

    Assertion `unifiedNumPos == (displsPos[Np - 1] + recvcntsPos[Np - 1])' failed.
    unifiedNumPos = 1
    (displsPos[Np - 1] + recvcntsPos[Np - 1]) = 49
    failed MPI.Gather() result at proc 0 (unifiedNumPos)
    queso/src/basic/src/VectorSequence.C, line 504, compiled Sep  1 2015 at 12:51:06
    
    terminate called after throwing an instance of 'QUESO::LogicError'
      what():  Error in QUESO internal logic
    Aborted (core dumped)

dmcdougall:
@briadam Is the example that generates that error pretty small/publicly viewable?  If so, would you mind sharing it?

dmcdougall:
> Dakota can't currently pass QUESO a comm of size > 1

What do you mean?  Do you get an error message?

briadam:
RE: comm of size > 1; no, no error; this is a Dakota limitation.  Dakota doesn't currently have a mechanism to allocate more than one MPI rank to run a nondeterministic method such as QUESO.  So when we run in MPI we're currently always handing you a comm of size = 1.  There are some solvers in Dakota that allow comm size > 1, just hasn't propagated to all yet.

dmcdougall:
Ah ok, thanks for clarifying.

briadam:
Reproducing the compile w/MPI, run in serial case would require you to run a (lightly) patched version of Dakota on any one of the 3 Dakota/QUESO test files in our repo.  I can try to gen that up if helpful, assuming you have a working Dakota + QUESO build environment.  I can also try to see if I can propose a patch based on a bit of debugging to see how we're getting into that call midway through in inverse problem solve.

dmcdougall:
Trying to find the Dakota/QUESO test files in the Dakota repository, but coming up thin.  I'm looking [here](https://software.sandia.gov/trac/dakota/browser/trunk#test).

briadam:
They are the files test/dakota_*queso.in.  I can either send you the relevant source code patches to invoke QUESO this way, or perhaps we should open an issue where we can attach files?

csim314:
Yes please open an issue. Issues are good and make baby Jesus happy.

dmcdougall:
On Tue, 1 Sep 2015, at 06:02 PM, Brian M. Adams wrote:
> They are the files test/dakota_*queso.in.  I can either send you the
> relevant source code patches to invoke QUESO this way, or perhaps we
> should open an issue where we can attach files?

Can't attach files to issues other than images, unfortunately.

dmcdougall:
@briadam Just leaving a public papertrail here with some more data points regarding when you can and can't reproduce that error.  Please correct me if any of this is wrong.

1.  The [error above](https://github.com/libqueso/queso/pull/393#issuecomment-136837987) can be triggered directly from the `test_serialEnv` test in our `make check` suite.
2.  The test passes on travis.  [Here's](https://github.com/libqueso/queso/blob/dev/.travis.yml) the `.travis.yml`.
3.  The test passes on my SL6 box with gcc/4.7, mpich2/1.4.1p1, boost/1.51.0, autotools/1.1, gsl/1.15
4.  The test passes on @briadam's Mac.
5.  The test fails on @briadam's RHEL6 machine with openmpi-1.8, boost-1.49, gsl-1.15, and gcc-4.4.7.

dmcdougall:
It's just occurred to me we don't configure with `--disable-mpi` on travis.  I'll try that now.

briadam:
With the help of valgrind I figured out what is going on.  To summarize, I have an MPI build of QUESO on RHEL6, openmpi-1.8, boost-1.49, gsl-1.15, and gcc-4.4.7.  If I run the test_serialEnv in serial, it fails.  

It's failing because of memory corruption in the "else" clauses in QUESO::MpiComm::Gather and Allreduce, which provide the override for cases where communication isn't needed.  The memcpy is either overreading or overwriting the bounds of the array, depending on the call.  These are called from VectorSequence.C, lines 79, 477, 491, and 518.  The passed arrays are of ints.

The root cause feels oddly familiar, but I can't recall why it happens: 4 = sizeof(int) != sizeof(MPI_INT) = 8.

If I change the data types of the following variables in VectorSequence.C to long int (likely not the right fix), it works: unifiedNumSamples, subNumSamples, subNumPos, auxBuf, unifiedNumPos, recvcntsPost, subNumObs, and recvcntsDbs.

dmcdougall:
Alright.  This is troublesome because I'm not sure how to get the size of MPI_INT without having assumed MPI_Init was called.

dmcdougall:
@briadam Just added a commit (trying something).  Can you try again?  We're going to have to back-and-forth for a little while because I can't reproduce the error locally or on travis.

dmcdougall:
Argh, of course I can't dereference a void pointer.  Hmm.

briadam:
Yeah, I don't see any easy out, without making the MpiComm interfaces strongly typed instead of void*.  It also makes be wonder why historical Dakota/QUESO uses cases work.  There we build with MPI and mpirun -n 1, so NumProc = 1 and I'd expect we'd overrun the buffers in the else clause.

dmcdougall:
The else clauses have only ever existed in this branch.  That's a good point actually.  Do you still get a test failure without --disable-mpi?

briadam:
Sorry for any confusion.  I'm only testing with MPI enabled (as I'm trying to test the case where we build with MPI, but run serial), and get the error whether mpirunning or serial running the test.  

For this case, I suspect that MpiComm.C:348 is to blame: dataTypeSize = sizeof(datatype).  For OpenMPI, I think MPI_INT is a pointer to a structure, so sizeof gives 8 on my 64-bit platform.  I don't see an easy workaround your switch for the non-MPI case relies on the fact that datatype has been typedefed.

dmcdougall:
> For this case, I suspect that MpiComm.C:348 is to blame: dataTypeSize = sizeof(datatype).

I came to the same conclusion.

dmcdougall:
@briadam I've been working with (read relentlessly bugging) @roystgnr about how they manage some of their MPI calls.  I've heavily stolen some code, would you mind giving it a try?  Tests pass locally for me.

Obviously this is only half-baked so far.  Just wanted to get something so you get a test to pass.  After you report back I can go through and do a more thorough update on the other communication methods as well.

briadam:
Yes! It works, though had to go a bit further to get the test clean.  I had to instantiate your Allreduce template on additional types int and unsigned int.  I then applied your new pattern to Gather and instantiated it on int.  I changed all but one of the MPI calls in BaseVectorSequence unifiedPositionsOfMaximum to use the new prototype, and it seems to pass.  Cool beans.

I also get a seg fault in test_uqGslVector, but presumably it's unrelated, so will dig into that after we get this resolved.

dmcdougall:
> I had to instantiate your Allreduce template on additional types int and unsigned int. I then applied your new pattern to Gather and instantiated it on int. I changed all but one of the MPI calls in BaseVectorSequence unifiedPositionsOfMaximum to use the new prototype, and it seems to pass. Cool beans.

This is the 'more thorough update' I was talking about in my previous comment.  Wasn't expecting you to have to do all that.

Glad it works.  I'll try to get this cleaned up today and add some docs for the new communicator methods.  Thanks for guinea-pigging.

dmcdougall:
Ok, docs are still outdated but the code is there now.

dmcdougall:
@briadam Does this pass unmodified for you now?

briadam:
Yes, this looks clean for RHEL6 in terms of an autoconf build of QUESO standalone (I still have one failing test #405, but it's unrelated.)  All the Dakota test cases are now passing as well:
 * Compile Dakota/QUESO with MPI, mpirun -n 1
 * Compile Dakota/QUESO with MPI, serial run
 * Compile Dakota/QUESO without MPI (serial run)
I'm game for merging this to libqueso:dev whenever you're happy with it.  I'll either pull that into Dakota, or wait for a release if one is imminent.

Thanks for the quick work on this!

dmcdougall:
Alright, thanks for checking this for me.  I'm going to spend the rest of the afternoon cleaning up this branch.  I'll be editing history so if you need the latest version be wary of that.

I'll ping again when this branch is cleaned up and ready to be merged.

dmcdougall:
Ok, this is clean now.  Each commit should pass `make check`.

dmcdougall:
If there's no further feedback by the end of the day I'll merge this (pending Travis checks).

Issue #388, opened 2015-06-17
v0.52: fp_compute{Covariances,Correlations}


beanSnippet:
If either ```fp_computeCovariances``` or ```fp_computeCorrelations``` is turned on in the input file, then I get this error:
```
Assertion `pqCorrMatrix(i,j) >= -1. - 1.e-8' failed.
pqCorrMatrix(i,j) = -nan
-1. - 1.e-8 = -1
computed correlation is out of range
basic/src/VectorSequence.C, line 2878
```
I get the same chains (for IP and FP) when I run with these options turned off.

dmcdougall:
The offending commit is 7725a3b5de362f5bf88bed0298628a26e41da6c3.

This looks like we missed some subtle logic issues when refactoring our assert macros.

dmcdougall:
Or perhaps not, since the correlation matrix has a nan element.

dmcdougall:
Ok, so the reason you now get an error is because @roystgnr's macro updates fixed a bug.  If you look at [this](https://github.com/libqueso/queso/commit/7725a3b5de362f5bf88bed0298628a26e41da6c3#diff-7a01fb53baade23ab8784706267b9864L2728) line you'll see that a `<` comparison is executed.  When a comparison is done on a `NaN`, the result is `false`.  Given the test is computed whether or not a value *is* between certain bounds, a result of `false` is interpreted as "`NaN` is between the acceptable bounds", which is the incorrect interpretation.

The new macros, however, reverse the test.  So now the test reads "reject everything not between these bounds."  Remember for `NaN`s, the comparison test will result in `false`, which is now the correct interpretation.

What's my point?  The new macros are superior to the old macros.  So this means some investigation needs to be done regarding where and why this function is running into a `NaN`.

dmcdougall:
And the reason you're getting a `NaN` here is because, from the example you emailed me, queso is trying to compute the correlations between things using a single posterior sample ([there is no check on the third argument to this function being 1](https://github.com/libqueso/queso/blob/dev/src/stats/src/StatisticalForwardProblem.C#L288)).  You need at least 2 samples to compute a variance.

Setting either `ip_mh_filteredChain_lag = 2` to get five samples, or `ip_mh_filteredChain_generate = 0` to use all ten samples fixes the problem.

That said, we can approach this two ways.  I can add a check to queso so that it either a) fails a little more gracefully when trying to compute the correlation with one sample; or b) add a check to skip the calculation (and not fail at all).

[...]

PR #394, opened 2015-08-21
Add some sanity checks to covariance computation

dmcdougall:
This adds some `queso_require_` checks to the
`ComputeCovCorrMatricesBetweenVectorSequences` method.  Specifically, we're
adding a check that we're computing a covariance/correlation using at least
two samples and we're also checking that the sample variance is positive in
every component before doing a division.

See #388 for discussion.


dmcdougall:
Does this seem like a reasonable solution?

dmcdougall:
Tests are passing and no objections were raised.  Merging.

PR #395, opened 2015-08-25
The term involving the determinant of \Sigma (the observational error…

JRigelo:
… covariance matrix) was added and test was adjusted accordingly


src/stats/src/GaussianLikelihoodFullCovarianceRandomCoefficient.C
```diff
@@ -81,9 +81,19 @@ GaussianLikelihoodFullCovarianceRandomCoefficient<V, M>::lnValue(
...
+  deter_cov = std::sqrt(deter_cov);
+
+  // Set the right hyperparameter coefficient
   // The last element of domainVector is the multiplicative coefficient of the
   // covariance matrix
-  return -0.5 * norm2_squared / (domainVector[domainVector.sizeLocal()-1]);
+  double cov_coeff = domainVector[domainVector.sizeLocal()-1];
+  cov_coeff = std::pow(std::sqrt(cov_coeff), domainVector.sizeGlobal()); 
+
+  return -0.5 * norm2_squared/ (domainVector[domainVector.sizeLocal()-1]) - std::log((cov_coeff*deter_cov));
```
dmcdougall: 1.  `deter_cov` is constant.  Is there a reason this is computed but 2 \pi^{d/2} is not computed?
2.  This is not crucial, but `domainVector.sizeGlobal()` is an integer.  `std::pow` can be computed more efficiently as a `for` loop if you wish.
3.  Taking `std::log` after a `std::sqrt` can be made more efficient by computing `0.5 * std::log`.

dmcdougall:
Replaced by #452.

PR #398, opened 2015-08-28
Add SharedPtr

dmcdougall:
In a similar vein to #368, this adds a `SharedPtr` type.


pbauman:
:+1:

roystgnr:
Well, it adds a SharedPtr<T>::Type type, anyway.  Damn C++03.

:+1:

roystgnr:
Silly thought: could/should we use inheritance here, rather than the nested type trick, so we don't have to write ::Type everywhere?

pbauman:
I defer to @roystgnr's sensibilities here.

PR #401, opened 2015-09-01
Compat gcc 4.4.7

briadam:
@dmcdougall Small change to allow compilation on gcc-4.4.7.  Not sure it's the right fix though, so request review.


roystgnr:
"Doesn't compile with a clang release" suggests "not the right fix", or at the very least "clang is broken in a way that will force a different workaround from us".  Let me take a look.

roystgnr:
Wait, no, looking at the Travis output it seems everything compiled but we had a regression test failure.

briadam:
Yeah I was surprised by that too.  I was also surprised to see typename there at all and it didn't seem that class or member needed a template parameter...

roystgnr:
And 0d96ae8 does look like the right fix to me.  You need to manually specify "typename" if (and only if, I thought, but the fact that newer compilers seem to be accepting the redundant typename specifier is a point against me) you're using a member type of a free-template-parameter-dependent class.  The outer class here isn't even templated, so there's no way that applies.  This is probably an atavism from the input parsing refactoring that happened a while back.

What is going on with 3fb937d, though?  Did you merge a zillion upstream changes at some point when you should have rebased?

Could you try making a new branch off dev, cherry-picking just 0d96ae8 onto that branch, and submitting that as a PR (either by making a new PR or by force-pushing to briadam:compat_gcc-4.4.7 to update this one)?  That'll kill two birds with one stone, by triggering a second Travis build so we can verify that the regression failure was just a fluke.

briadam:
Yes, I thought I did the merge from upstream wrong.  I tried to do it with the Github web interface and probably picked the wrong options.  I swear I'm trying to be less of a Git idiot, but not making much progress.

dmcdougall:
> I swear I'm trying to be less of a Git idiot, but not making much progress.

If it makes you feel better, the mistake I learned most from was rebasing the `master` branch of matplotlib.  I needed a change of underwear once I realised what I had done (and that I also had write access to the upstream repository).

Luckily, I didn't `push -f` so it was just the local branch I nuked.  No harm done.

roystgnr:
It's not the smoothest learning curve, is it?  I used Git for months before I was comfortable doing everything in branches, and months more before I realized that I didn't have to carefully keep my fork's master in sync with origin/master (as I'm guessing you tried to do with that commit?), I could just branch a new roystgnr/whatever directly from origin/master.

briadam:
Okay, learned a few things.  Closing this one.

PR #403, opened 2015-09-02
Inline GSLMatrix::operator()

roystgnr:
And trim some of the redundant non-disablable range tests while we're
at it.

Matrix access isn't as pervasive as Vector, but it's still too
inner-loopy to want a function call and non-debug-mode range checking
on every entry.

I probably should have thought to check this when I did #391, not just upon
"I'm writing code to copy out a submatrix; wonder how fast that's going to be..."


dmcdougall:
Pending Travis checks, this looks good.  Thanks.

PR #404, opened 2015-09-02
Implement sampler interface base class

dmcdougall:
This is a work in progress.  Multilevel not quite there yet, but the interface is there.  It's likely to change but any initial feedback is greatly appreciated.

The idea here is that we'd like to create other sequence generators (`MetropolisHastingsSG` is a sequence generator for the DRAM algorithm).  Robust sequence generators (samplers) in infinite dimensions, derivative-informed samplers, and even user-provided samplers, are all on the radar here.


dmcdougall:
Also note that the docs for the base class are totally wrong.  They're copy-pasted from `MetropolisHastingsSG`.  I'll update them in due course.

dmcdougall:
Hmm, that's weird.  The tests passed locally for me.

dmcdougall:
[This](https://github.com/libqueso/queso/pull/404/files#diff-3e199cbdc934b269a2f017e7589f59c7) is an example of creating a custom sequence generator and passing it to the `StatisticalInverseProblem`.  It isn't the cleanest, but it's a good initial attempt.

Still left to do is to clean up some of the memory leaks I've inevitably introduced here; to make only the `generateSequence` method pure virtual; to implement an actual algorithm by creating a custom transition kernel; and to clean up the base class documentation.

dmcdougall:
After thinking about this a little more, I'm not sure this is the right approach.  With this design, the user would subclass `SequenceGenerator` and implement `generateSequence`, which requires the user to generate the whole chain, including all of the accept/reject logic.  I think that's asking for too much from the user.

I think a better approach would be to ask the user to implement a different pure virtual method that provides the current state, the current iteration, and returns the proposed state.  This would mean that queso still does all of the heavy lifting for the accept/reject and all the user has to do is provide the logic for transitioning.

roystgnr:
Can you split the difference?  Create an intermediate pure virtual subclass wherein QUESO implements generateSequence but then gives the user the easier pure virtual method to do the case-dependent work?

dmcdougall:
Yeah, I think that's where I was going but hadn't nailed down the hierarchy in my head.  That's definitely what I'm going to try now.  The good news is that with your suggestion, the work done here can be reused.

dmcdougall:
Ok I think most of the machinery is here now.  The [test](https://github.com/libqueso/queso/pull/404/files#diff-9b0c7c1db769c7fe7e983d50f6c733b9) I implement just runs a custom sampler (vanilla random walk) on a 2D Gaussian with mean 10 and covariance the identity matrix.

[Here's](https://github.com/libqueso/queso/files/2784/samples.pdf) example output from this.

There's still a lot of odds and ends to clean up, but I figured I'd leave an update to keep everyone apprised.

dmcdougall:
Also the 'test' is not really a test; I need to actually compute test statistics on the output.

dmcdougall:
There exists a test.

Issue #405, opened 2015-09-03
uqGslVector test seg fault


briadam:
On my RHEL6 build, test_uqGslVector seg faults because the env is deleted before the destructors for the VectorSpace objects are called (and the VS destructor refers to env).  This can be fixed by deleting param_space before deleting env, and scoping big_param_space so it gets destroyed before the env is deleted.  I didn't propose a patch because I wasn't sure how you'd prefer to fix this...

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7bd835f in QUESO::BaseEnvironment::displayVerbosity (this=0x86d9d0)
    at ../../src/core/src/Environment.C:398
398	  queso_require_msg(m_optionsObj, "m_optionsObj variable is NULL");

#0  0x00007ffff7bd835f in QUESO::BaseEnvironment::displayVerbosity (this=0x86d9d0)
    at ../../src/core/src/Environment.C:398
#1  0x00007ffff7c5b8f0 in QUESO::VectorSpace<QUESO::GslVector, QUESO::GslMatrix>::~VectorSpace (this=0x7fffffffd910, __in_chrg=<value optimized out>)
    at ../../src/basic/src/VectorSpace.C:133
#2  0x000000000040a11c in main ()

dmcdougall:
Fixed in #406.

PR #406, opened 2015-09-11
Fix uqGslVector test seg fault

briadam:
Proposed fix for #405

### DIFF
--- a/test/test_GslVector/test_uqGslVector.C
+++ b/test/test_GslVector/test_uqGslVector.C
@@ -79,9 +79,9 @@ int main(int argc, char **argv) {
   }
 
   // Testing concatenate so we need a bigger param space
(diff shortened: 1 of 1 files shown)


pbauman:
Was the problem just that the env was being deleted before the parameter space was destroyed? Smaller fix would just be to scope the parameter space, right? Regardless, thanks for reporting and taking a crack at the fix.

briadam:
Yes, you are correct.  I first fixed it that way (adding braces) for big_param_space, but then realized that param_space was constructed via new and would have to be deleted anyway (or similarly scope-limited), so just decided to follow the existing memory management pattern rather than have two.

dmcdougall:
I wrote this test before I knew about safer memory manage patterns and smart pointers.  I'm ok with this fix, although in future I think passing references or `SharedPtr`s around would be preferable.

pbauman:
:+1:

dmcdougall:
@briadam Thanks!

PR #413, opened 2015-12-01
Fix compatibility with recent libMesh

roystgnr:
This is the same bug as was fixed in 25fd4c51 - I just missed this
instance because I hadn't been using a SLEPc-enabled libMesh build.


dmcdougall:
+1

PR #414, opened 2015-12-16
Update to newest https://github.com/tsuna/boost.m4

roystgnr:
Using https://github.com/tsuna/boost.m4/commit/df28a91500c3263662868f3bc8912b2ed016972b fixes configuration for me on Ubuntu 15.10


dmcdougall:
I see `gcc-5.x.x`-specific foo in there, so I'm happy that's now supported.  Also, just as a note to myself and to leave a paper trail from a private conversation, this new `m4` file [removes support for reading `$BOOST_DIR` at configure time](https://github.com/libqueso/queso/pull/414/files#diff-2953ba2c8059e513050cbb6b1cd16fadL110).  I'm okay with this but if of our users scream loudly enough we might want to consider dropping that back in.

PR #418, opened 2016-02-04
Fix sign error in lnValue of JeffreysJointPdf

dmcdougall:
The Jeffreys prior is 1.0 / x.  Log of it is log(1.0 / x) = -x, but what was returned was the negative of this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -214,3 +214,4 @@ test/test_SharedPtr
 test/output_test_serialEnv
 test/test_Environment/input_test_serialEnv
 test/test_serialEnv
+test/test_jeffreys_pdf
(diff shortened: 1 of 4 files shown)


dmcdougall:
> The Jeffreys prior is 1.0 / x. Log of it is log(1.0 / x) = -x, but what was returned was the negative of this.

Oops, I meant log(1.0 / x) = -log(x).

beanSnippet:
I am confused-- I thought the pdf was supposed to transform a sample from a uniform distribution (domainVector[i]) to another distribution, not evaluate the pdf at that point...?

dmcdougall:
Are you referring to [Inverse Transform Sampling](https://en.wikipedia.org/wiki/Inverse_transform_sampling)?  `actualValue` at a point should certainly return the pdf of the random at that point.

beanSnippet:
Oh, I think I was thinking of ```VectorRealizer```. Nevermind.

PR #420, opened 2016-02-12
Add gradients of Gaussian distribution (with tests)

dmcdougall:


dmcdougall:
@nicholasmalaya Do you mind checking this out?

dmcdougall:
@nicholasmalaya Added another analytical test for a non-zero gradient.  Let me know what you think.

src/stats/src/GaussianJointPdf.C
```diff
@@ -139,17 +139,23 @@ GaussianJointPdf<V,M>::actualValue(
 
   queso_require_equal_to_msg(domainVector.sizeLocal(), this->m_domainSet.vectorSpace().dimLocal(), "invalid input");
 
-  queso_require_msg(!(gradVector || hessianMatrix || hessianEffect), "incomplete code for gradVector, hessianMatrix and hessianEffect calculations");
+  queso_require_msg(!(hessianMatrix || hessianEffect), "incomplete code for gradVector, hessianMatrix and hessianEffect calculations");
 
   double returnValue = 0.;
 
-  if (this->m_domainSet.contains(domainVector) == false) { // prudenci 2011-Oct-04
+  if (this->m_domainSet.contains(domainVector) == false) {
```
nicholasmalaya: What is this logical testing for? Is it just testing if there are no model parameters for this pdf? If that is the case, does one want to return 0, instead of an error?

test/test_pdfs/test_gaussian_pdf_gradient.C
```diff
@@ -0,0 +1,112 @@
...
+  var(2,2) = 7.0;
+
+  // Check the zero-gradient case
+  QUESO::GaussianJointPdf<> pdf("", paramDomain, mean, var);
+
+  QUESO::GslVector lnGradVector(paramSpace.zeroVector());
+  QUESO::GslVector gradVector(paramSpace.zeroVector());
+  QUESO::GslVector point(mean);
+
+  double lnPdfValue1 = pdf.lnValue(point, NULL, &lnGradVector, NULL, NULL);
```
nicholasmalaya: Before running this test, does it make sense to test the case where you do not explicitly pass &lnGradVector (instead, pass NULL)? In this case, you would only be evaluating the value of the pdf, not the gradient. I don't see this being tested anywhere else at present but its an easy edge case and not everyone wants to perform gradient evaluations. Just a thought.

src/stats/src/GaussianJointPdf.C
dmcdougall: No, it's testing whether or not `domainVector`, the point we're evaluating the pdf at, is actually inside the domain or not.  The user may have constructed a bounded domain.  In this case, the pdf is zero (no probability).

test/test_pdfs/test_gaussian_pdf_gradient.C
dmcdougall: It's tested [here](https://github.com/libqueso/queso/blob/dev/test/test_GaussianVectorRVClass/test_VectorPdf_gsl.C#L54).

dmcdougall:
@nicholasmalaya If Travis comes back happy, is this good to go?

dmcdougall:
@nicholasmalaya ping

nicholasmalaya:
@dmcdougall Looks much improved to me, I don't have any outstanding issues. Thanks for the effort. I'm firing torpedos and merging.

PR #425, opened 2016-02-17
Gpmsa multivariate

roystgnr:
Not yet certain or confident this is working, but it's at least passing internal assertions, still passing the scalar regression test (after one change in gold standard data that was carefully tested to correspond to a bugfix in the scalar case), giving numbers instead of NaNs, and basically ready for outside review.


src/core/src/GslMatrix.C
```diff
@@ -692,15 +692,17 @@ GslMatrix::filterLargeValues(double thresholdValue)
...
+  const MpiComm & comm = this->map().Comm();
+  Map serial_map(nCols, 0, comm);
+
+  GslMatrix mat(m_env,serial_map,nRows);
 
-  GslMatrix mat(m_env,m_map,nCols);
   for (unsigned int row = 0; row < nRows; ++row) {
     for (unsigned int col = 0; col < nCols; ++col) {
-      mat(row,col) = (*this)(col,row);
+      mat(col,row) = (*this)(row,col);
```
dmcdougall: Is there a performance motivation underlying this change?

src/core/inc/GslMatrix.h
```diff
@@ -197,6 +197,17 @@ class GslMatrix : public Matrix
   //! This function multiplies \c this matrix by vector \c x and returns the resulting vector.
   GslVector  multiply                  (const GslVector& x) const;
 
+  //! This function multiplies \c this matrix by vector \c x and fills
+  //the preallocated vector \c y.
+  void       multiply                  (const GslVector& x, GslVector& y) const;
+
+  //! This function multiplies \c this matrix by matrix \c X and returns the resulting matrix.
+  GslMatrix  multiply                  (const GslMatrix& X) const;
```
dmcdougall: This may be fine, but for high-dimensional problems this returns an expensive copy.  Other alternatives are to fill a preallocated `GslMatrix`, or a non-`const` method that just updates `*this`.

src/gp/inc/GPMSA.h
```diff
@@ -288,78 +339,116 @@ class GPMSAFactory
...
+  typename ScopedPtr<V>::Type m_discrepancyPrecisionShapeVec;
+  typename ScopedPtr<V>::Type m_discrepancyPrecisionScaleVec;
+  typename ScopedPtr<V>::Type m_discrepancyCorrelationStrengthAlphaVec;
+  typename ScopedPtr<V>::Type m_discrepancyCorrelationStrengthBetaVec;
+  typename ScopedPtr<V>::Type m_emulatorDataPrecisionShapeVec;
+  typename ScopedPtr<V>::Type m_emulatorDataPrecisionScaleVec;
 
   // The gaussian process object to build
-  GPMSAEmulator<V, M> * gpmsaEmulator;
+  typename ScopedPtr<GPMSAEmulator<V, M> >::Type gpmsaEmulator;
```
dmcdougall: Ugh, thanks for doing all that.  I think when I initially implemented this we didn't have the `ScopedPtr` foo.

src/core/src/GslMatrix.C
roystgnr: The other lines were to get transposes working with square matrices.  This line was changed to keep the rectangular dimensions consistent with the definitions above.

I didn't think about the possibility of performance change.  I don't think there's any clearly-winning way to do a transpose, since if you're doing a nice stride on one matrix you're doing a poor stride on the other, but we could swap the loop order if I'm wrong.

src/core/inc/GslMatrix.h
roystgnr: Yeah, this was on my list of "stuff to optimize after getting correct results", but since we don't end up calling this function from within the likelihood, only from within the setup, I'm no longer planning to bother.

src/gp/src/GPMSA.C
```diff
@@ -497,6 +693,195 @@ GPMSAFactory<V, M>::addSimulations(
...
+      this->m_simulationOutputs,
+      this->m_experimentScenarios,
+      this->m_experimentOutputs,
+      this->m_discrepancyBases,
+      this->m_observationErrorMatrices,
+      *(this->m_experimentErrors),
+      *(this->m_totalPrior),
+      *this->residual,
+      *this->BT_Wy_B_inv,
+      *this->KT_K_inv));
```
dmcdougall: Note to self, the GPMSAEmulator ABI changed (new ctor arguments added), but it's encapsulated by the GPMSAFactory.  Furthermore, although these things are not needed in the scalar case, they are still calculated but they are just ignored.

src/core/src/GslMatrix.C
dmcdougall: Ah, got it.  Thanks.

dmcdougall:
@roystgnr Shall we go ahead and merge this?  It's targeting `dev` so won't go into `v0.55.0-release`, giving us more time to pursue verification efforts.

roystgnr:
Okay, *now* we can merge this, once Travis signs off.  We do still need an actual verification test, but for now I'll settle for a regression test.

roystgnr:
Looks like this is so old it missed the QUESO_CPPFLAGS/AM_CPPFLAGS build system refactoring.  So git *thinks* the merge can be done cleanly but that's a lie.  I'll rebase and fix it now.

PR #426, opened 2016-03-04
Fix issue involving unified MLE on non-inter0Comm

dmcdougall:
This patch adds some safeguards to avoid code that does inter0Comm computing
from being executed by processes that don't belong to the inter0Comm.
Specifically, computing the unified MLE (and positions) and computing
correlation matrices are to only be done on the inter0Comm, since this is where
all the samples live anyway.

Marked as 'do not merge' because this still needs a test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -218,3 +218,9 @@ test/test_jeffreys_pdf
 test/test_gaussian_pdf_gradient
 test/test_log_normal_pdf_gradient
 test/test_beta_pdf_gradient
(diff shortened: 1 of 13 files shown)


dmcdougall:
I guess as a test I can run the gravity example with:

1.  One subenvironment, one processor
2.  One subenvironment, two processors

And check the output is the same.

dmcdougall:
Ok, test added.  If this passes, I'll merge it.  It passes locally with and without the `--disable-mpi` configure flag.

PR #429, opened 2016-03-26
Support hdf5 output

dmcdougall:
Specifying, for example, `ip_mh_rawChain_dataOutputFileType = h5`, will write the raw chain to an HDF5 file.

The code here can definitely use some optimisation, but it's a good first start.

### DIFF
--- a/src/core/src/Environment.C
+++ b/src/core/src/Environment.C
@@ -600,9 +600,22 @@ BaseEnvironment::openOutputFile(
           filePtrSet.ofsVar = new std::ofstream((baseFileName+"_sub"+this->subIdString()+"."+fileType).c_str(),
                                                 std::ofstream::out /*| std::ofstream::in*/ | std::ofstream::app);
         }
+#ifdef QUESO_HAS_HDF5
         else if (fileType == UQ_FILE_EXTENSION_FOR_HDF_FORMAT) {
-          queso_error_msg("hdf file type not supported yet");
+          std::string fullFileName =
+            baseFileName+"_sub"+this->subIdString()+"."+fileType;
+
+          // Use H5F_ACC_EXCL because not overwriting, so fail on existing file
+          filePtrSet.h5Var = H5Fcreate(fullFileName.c_str(),
+                                       H5F_ACC_EXCL,
+                                       H5P_DEFAULT,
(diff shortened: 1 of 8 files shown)


src/core/src/Environment.C
```diff
@@ -112,6 +112,10 @@ FilePtrSetStruct::FilePtrSetStruct()
   :
   ofsVar(NULL),
   ifsVar(NULL)
+#ifdef QUESO_HAS_HDF5
+  ,  // lol
```
nicholasmalaya: ? 

What does this do?

nicholasmalaya:
I'm not seeing a regression test that just touches this capability to ensure that it can successfully write foo.h5 -- am I wrong?

If no, I can probably add one, should be trivial. Just want to check I'm not missing something

src/core/src/Environment.C
dmcdougall: If QUESO has HDF5, this code is compiled.  Without the comma, the previous line does not have a comma and you get a compiler error.  If QUESO does not have HDF5, the previous line is the last line in the initialisation list and therefore cannot have a comma.

"Lol" means "laughing out loud", and it is there because a single line with a comma amused me greatly.

dmcdougall:
> I'm not seeing a regression test that just touches this capability to ensure that it can successfully write foo.h5 -- am I wrong?

You are not wrong.  I am taking suggestions for tests.  A successful write is a good idea.  Would a better idea be to write both HDF5 output and ASCII output and check that they're both the same?

nicholasmalaya:
A successful write is necessary, I don't think comparing ASCII and HDF is critical since it would require a relative diff. Just writing a vector of 1s would probably be enough to merge.

dmcdougall:
@nicholasmalaya Done!

dmcdougall:
I didn't `#ifdef` guard the test.  I'll fix that.

dmcdougall:
@nicholasmalaya Ok done, let me know what you think.

nicholasmalaya:
If you append something like, 'Support HDF5 output' to the changelog I'm game to merge.

dmcdougall:
@nicholasmalaya done

nicholasmalaya:
:+1:

PR #439, opened 2016-04-13
Refactor delayed rejection into a method

dmcdougall:
Tests pass locally.


dmcdougall:
Ah, I should write better documentation for this method.

dmcdougall:
Ok I think this is good to go now.

PR #440, opened 2016-04-15
Fix broken test on case-insensitive filesystems

dmcdougall:
Inside test/ there is a directory called test_GPMSA_vector/.
test_GPMSA_vector/test_gpmsa_vector.C gets used to build a binary called
test_gpmsa_vector.

In short, after configuring, `make clean` fails because while test_gpmsa_vector
is a clean target, on a case-insensitive filesystem that pattern matches
test_GPMSA_vector.  The build system then rightfully complains that it can't
possibly clean up the (nonexistent) test because it is a directory.

The fix is just to rename the test directory.


dmcdougall:
@roystgnr Look good?

roystgnr:
Great; thanks!

PR #430, opened 2016-03-29
std::isnan, std::isfinite, std::unique_ptr, std::shared_ptr

pbauman:
Now, C++11 is turned on by default in the compiler. Adds infrastructure to test for, and use if detected, std::isnan, std::isfinite, std::unique_ptr, and std::shared_ptr; these tests were inspired by or straight up copied from libMesh/cxx11.m4. We fall back to Boost if not supported. @roystgnr note that I had to turn off the C++11 style asserts in 5cb0940 to get QUESO to compile with C++11. There was a cascading set of failures to convert types. I tried to replace all the places we use Boost with the macros/typedefs, except for program options. So one step in the direction of making Boost optional. @dmcdougall I didn't have a SLEPc-enabled libMesh lying around, so I may have missed something in there. Could you please check that I didn't hose that?

I'm not sure we want to merge this right away. I like moving farther away from the boost requirement, but program_options uses boost::shared_ptr under the hood and so there are LOTS of warnings emitted about std::auto_ptr deprecation (since boost::shared_ptr provides an interface to std::auto_ptr). Those warnings definitely are verbose and could hide legit warnings.

PR #441, opened 2016-04-15
Use boost::math::isnan, don't rely on C99/C++11

roystgnr:
Hopefully this will fix the problem @dmcdougall hit on OSX.

### DIFF
--- a/src/gp/src/GPMSA.C
+++ b/src/gp/src/GPMSA.C
@@ -26,6 +26,8 @@
 #include <queso/GslVector.h>
 #include <queso/GslMatrix.h>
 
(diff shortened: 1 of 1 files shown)


pbauman:
Would be nice to use the solution in #430

pbauman:
Which is not yet merged, I should say.

roystgnr:
Shit.  I should try to keep up.

roystgnr:
The goal is for boost to be an optional dependency whenever we have C++11?  In that case #430 is definitely the better solution.

dmcdougall:
Let's use this as an intermediate patch before #430.

dmcdougall:
Thanks @roystgnr.

PR #442, opened 2016-04-20
Update documentation for Gamma distribution

dmcdougall:
GammaJointPdf had basically no documentation regarding what role the parameters
a and b of the distribution mean.  This patch adds docs for these.  This patch
also updates GammaVectorRealizer to add references to GammaJointPdf to guide
the user more easily through how the Gamma(a, b) distribution is defined.

### DIFF
--- a/src/stats/inc/GammaJointPdf.h
+++ b/src/stats/inc/GammaJointPdf.h
@@ -68,6 +68,10 @@ public:
    * Indeed, \c a and \c b are vectors and in multiple dimensions the pdf is
    * just the product of Gamma(a_i, b_i) pdfs distributions in each dimension.
    * That is, they are independent.
+   *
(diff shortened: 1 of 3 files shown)


dmcdougall:
@roystgnr Does this look better?

roystgnr:
Definitely an improvement.  Can we put in a note about how these are the k and theta parameters from https://en.wikipedia.org/wiki/Gamma_distribution ?  I think the source of my confusion is that my memory tries to hold "a means k, b means theta", but "a means alpha, b means beta" sounds so much more plausible that I never trust my own memory.

I don't think we should *change* the a, b syntax (it matches GSL, for one reason) but a note would be helpful.

roystgnr:
:+1:

dmcdougall:
Thanks.

PR #443, opened 2016-04-25
PDF means

roystgnr:
This adds a centroid() method for sets and a distributionMean() method for PDFs.  Using the latter on a prior distribution ought to be useful for automatically choosing good initial iterates in inverse problems.


dmcdougall:
I think this looks good.  Would you mind adding some tests for the means that don't throw exceptions?

roystgnr:
"+226,-4"

roystgnr:
"+1083,-4"

src/stats/src/ConcatenatedJointPdf.C
```diff
@@ -183,6 +183,24 @@ ConcatenatedJointPdf<V,M>::lnValue(
...
+  std::vector<V*> vecs(m_densities.size(),(V*) NULL);
+  std::vector<double> values(m_densities.size(),0.);
+  double returnValue = 0.;
+  unsigned int cumulativeSize = 0;
+  for (unsigned int i = 0; i < vecs.size(); ++i) {
+    vecs[i] = new V(m_densities[i]->domainSet().vectorSpace().zeroVector());
+    m_densities[i]->distributionMean(*(vecs[i]));
+    meanVector.cwSet(cumulativeSize,*(vecs[i]));
+    cumulativeSize += vecs[i]->sizeLocal();
+    delete vecs[i];
```
dmcdougall: You might consider using a shared pointer for these.

src/basic/inc/BoxSubset.h
```diff
@@ -66,6 +66,9 @@ class BoxSubset : public VectorSubset<V,M> {
    * m_minValues, and 2) all all components in \c vec are smaller than m_maxValues. */
   bool contains (const V& vec)     const;
 
+  //! Returns the centroid of this box subset.
```
dmcdougall: Not sure how precise you want to be in these doxygen comments.  The method doesn't actually return anything, it fills up `vec`.

src/stats/src/ConcatenatedJointPdf.C
roystgnr: Yeah, this is lousy code.  I was trying to follow the style in a couple of the other ConcatenatedJointPdf functions but I can just fix those instead.  :-)

dmcdougall: Don't worry about fixing up other parts of the file.  I can open a ticket for it so it's not forgotten.

PR #445, opened 2016-06-16
Fix a race condition in gsl test output directory

dmcdougall:
The output isn't even needed, so this patch just turns it off.


pbauman:
👍 That probably dates back to when I added those things. Thanks.

dmcdougall:
No worries.  Merging.

PR #451, opened 2016-07-19
Make boost program options optional

dmcdougall:
`configure --disable-boost-program-options`.

This is initial work, and the history certainly needs to be cleaned up before merging.  I'm opening this PR because I finally got the tests passing.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -84,6 +84,16 @@ AC_ARG_ENABLE(mpi,
               [enable_mpi=no],
               [enable_mpi=yes])
 
+##############################################
+# Check if the user disabled program options #
+##############################################
+AC_ARG_ENABLE([boost-program-options],
+              AS_HELP_STRING([--disable-boost-program-options], [Disables use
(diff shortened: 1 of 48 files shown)


dmcdougall:
Ok, I think I'm happy with this now.  Feedback welcome.

pbauman:
OK, so the strategy is you first have to opt-in to opt-out of Boost PO and then, later, we'll make opting out default?

dmcdougall:
Yeah, boost program options is enabled by default.  Passing in `--disable-boost-program-options` disables it and uses GetPot instead.

With #430 merged, I wonder if we should actually just use `--without-boost` instead, though.

dmcdougall:
I messed up something, apparently.

dmcdougall:
Oh yeah, ok, I removed headers instead of `#ifdef`ing around them.  Duh.

dmcdougall:
Ok we're good.

dmcdougall:
@pbauman Any comments before I merge this sucker?

PR #452, opened 2016-07-20
Fix normalising constant hyperparameter in canned likelihoods

dmcdougall:
Replaces #395.

### DIFF
--- a/src/stats/inc/GaussianLikelihoodFullCovarianceRandomCoefficient.h
+++ b/src/stats/inc/GaussianLikelihoodFullCovarianceRandomCoefficient.h
@@ -36,7 +36,12 @@ class GslMatrix;
  * \file GaussianLikelihoodFullCovarianceRandomCoefficient.h
  *
  * \class GaussianLikelihoodFullCovarianceRandomCoefficient
- * \brief A class that represents a Gaussian likelihood with full covariance and random coefficient
+ * \brief A class that represents a Gaussian likelihood with full covariance
+ * and random coefficient.
+ *
+ * The random coefficient is a scalar that pre-multiplies the covariance
+ * matrix.  This is treated as a hyperparameter to be inferred during
+ * the sampling procedure.
  */
 
 template <class V = GslVector, class M = GslMatrix>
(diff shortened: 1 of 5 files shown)


test/test_gaussian_likelihoods/test_fullCovarianceRandomCoefficient.C
```diff
@@ -129,7 +129,7 @@ int main(int argc, char ** argv) {
   point[0] = -2.0;
   point[1] = 1.0;
   lhood_value = lhood.actualValue(point, NULL, NULL, NULL, NULL);
-  truth_value = 1.0;
+  truth_value = std::exp(-std::log(2.0));
```
pbauman: Why this and not `1.0/2.0`?

src/stats/inc/GaussianLikelihoodFullCovarianceRandomCoefficient.h
```diff
@@ -36,7 +36,9 @@ class GslMatrix;
  * \file GaussianLikelihoodFullCovarianceRandomCoefficient.h
  *
  * \class GaussianLikelihoodFullCovarianceRandomCoefficient
- * \brief A class that represents a Gaussian likelihood with full covariance and random coefficient
+ * \brief A class that represents a Gaussian likelihood with full covariance
+ * and random coefficient, treated as a hyperparameter to be inferred during
```
pbauman: Is every coefficient in the matrix being treated as an independent parameter? If so, I might spell that out explicitly.

test/test_gaussian_likelihoods/test_fullCovarianceRandomCoefficient.C
dmcdougall: Because I copied the code from #395 without thinking about it.  I'll update it.

src/stats/inc/GaussianLikelihoodFullCovarianceRandomCoefficient.h
dmcdougall: No.  I'll clarify this.

PR #459, opened 2016-09-15
Converting asserts don't work for many data types

roystgnr:
These were designed for numeric comparisons, but try to do something
as innocuous as compare std::string to const char * and they fail with
compiler errors.  Best to remove them entirely for now.


pbauman:
Perhaps the better thing to do is rollback 2f72cc903f574132c46f1a44cc7aa2514e3ffddb if we want to do this? Personally, I'm fine throwing the extra `std::string()` in there if it adds some safety for the numeric types. Just my two cents.

roystgnr:
Nah; IIRC when I thought about it it didn't do much good for inter-type numeric comparisons at all.  IIRC the conversions would avoid a few false negatives in signed-vs-unsigned comparisons, but could theoretically add false positives to certain integer-vs-float comparisons.

PR #466, opened 2016-09-28
Implement a factory pattern for transition kernels

dmcdougall:
Allows the user to implement a custom transition kernel and register with the factory so QUESO can instantiate the custom kernel at run-time.

All commits pass `make check` locally, the changelog has been updated and documentation included.

This should replace #404 and #408.

### DIFF
--- a/src/stats/src/MetropolisHastingsSG.C
+++ b/src/stats/src/MetropolisHastingsSG.C
@@ -36,7 +36,7 @@
 #include <queso/GaussianJointPdf.h>
 
 #include <queso/TransitionKernelFactory.h>
-#include <queso/AcceptanceRatio.h>
+#include <queso/AlgorithmFactory.h>
 
 namespace QUESO {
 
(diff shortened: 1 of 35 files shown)


src/stats/src/MetropolisHastingsSG.C
```diff
@@ -514,7 +514,7 @@ MetropolisHastingsSG<P_V,P_M>::commonConstructor()
     msg = "The doLogitTransform option is deprecated.  ";
     msg += "Use ip_mh_algorithm instead.";
     std::cerr << msg << std::endl;
-    queso_warning(msg.c_str());
+    queso_do_once(queso_warning(msg.c_str()));
```
roystgnr: queso_warning already is defined to use queso_do_once

src/stats/src/MetropolisHastingsSG.C
```diff
@@ -513,7 +513,6 @@ MetropolisHastingsSG<P_V,P_M>::commonConstructor()
     std::string msg;
     msg = "The doLogitTransform option is deprecated.  ";
     msg += "Use ip_mh_algorithm instead.";
-    std::cerr << msg << std::endl;
```
roystgnr: How were you *not* able to replicate this locally?

src/stats/src/MetropolisHastingsSG.C
dmcdougall: Move along.  Nothing to see here.

src/stats/src/MetropolisHastingsSG.C
dmcdougall: Move along. Nothing to see here.

dmcdougall:
Ok we're cooking on gas now.

PR #471, opened 2016-10-09
Compartmentalise getpot

dmcdougall:
This should be fairly self-explanatory.

### DIFF
--- a/src/contrib/getpot/getpot.h
+++ b/src/contrib/getpot/getpot.h
@@ -134,8 +134,13 @@ typedef  std::vector<std::string>  STRING_VECTOR;
   std::vector<TYPE>::const_iterator ITERATOR = (VARIABLE).begin();      \
   for (; (ITERATOR) != (VARIABLE).end(); ++(ITERATOR))
 
-// Just put GetPot in the QUESO namespace.
-namespace QUESO {
+// We allow GETPOT_NAMESPACE to be defined before this file is
+// included; if libraries using two different versions of GetPot might
+// be linked together, the result may be unsafe unless they're put in
+// different namespaces.
+#ifdef GETPOT_NAMESPACE
+namespace GETPOT_NAMESPACE {
+#endif
 
 /**
  * GetPot - A class for parsing comand line arguments and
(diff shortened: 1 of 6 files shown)


roystgnr:
"In the matter of reforming things, as distinct from deforming them, there is one plain and simple principle; a principle which will probably be called a paradox. There exists in such a case a certain institution or law; let us say, for the sake of simplicity, a fence or gate erected across a road. The more modern type of reformer goes gaily up to it and says, “I don’t see the use of this; let us clear it away.” To which the more intelligent type of reformer will do well to answer: “If you don’t see the use of it, I certainly won’t let you clear it away. Go away and think. Then, when you can come back and tell me that you do see the use of it, I may allow you to destroy it.”
 
This paradox rests on the most elementary common sense. The gate or fence did not grow there. It was not set up by somnambulists who built it in their sleep. It is highly improbable that it was put there by escaped lunatics who were for some reason loose in the street. Some person had some reason for thinking it would be a good thing for somebody. And until we know what the reason was, we really cannot judge whether the reason was reasonable. It is extremely probable that we have overlooked some whole aspect of the question, if something set up by human beings like ourselves seems to be entirely meaningless and mysterious. There are reformers who get over this difficulty by assuming that all their fathers were fools; but if that be so, we can only say that folly appears to be a hereditary disease. But the truth is that nobody has any business to destroy a social institution until he has really seen it as an historical institution. If he knows how it arose, and what purposes it was supposed to serve, he may really be able to say that they were bad purposes, or that they have since become bad purposes, or that they are purposes which are no longer served. But if he simply stares at the thing as a senseless monstrosity that has somehow sprung up in his path, it is he and not the traditionalist who is suffering from an illusion."

roystgnr:
Or to move from the meta level to specifics:

Adding "#define GETPOT_NAMESPACE QUESO" in our config header would have had the same desirable effect, but without the undesirable side effect of forking getpot.h

dmcdougall:
Point taken.

Better?

pbauman:
I see @roystgnr was feeling saucy.

roystgnr:
:+1:, thanks!

PR #473, opened 2016-11-01
Make transition kernels user-updatable

dmcdougall:
`BaseTKGroup` learned the `set_current_iteration` method, which sets the internal `m_currentIteration` variable and calls the private `virtual` method `update_tk`.  The default behaviour of this method is to do nothing and the user can subclass this method to implement any iteration-dependent transition kernel behaviour.

I think, before merging this, it's worth discussing how this should interact with the adaptively update proposal covariance matrix and the delayed rejection functionality.

### DIFF
--- a/examples/Makefile.am
+++ b/examples/Makefile.am
@@ -515,6 +515,25 @@ example_SharedPtr_CPPFLAGS = $(AM_CPPFLAGS)
 dist_pointers_DATA =
 dist_pointers_DATA += ${example_SharedPtr_SOURCES}
 
+####################################
+# Custom Transition Kernel Example #
+####################################
+custom_tkdir = $(prefix)/examples/custom_tk
+
+custom_tk_PROGRAMS = hessian_eg
+
+hessian_eg_SOURCES =
+hessian_eg_SOURCES += custom_tk/hessian_eg.C
+hessian_eg_SOURCES += custom_tk/hessian_eg_main.C
+custom_tk_HEADERS = custom_tk/hessian_eg.h
+hessian_eg_LDADD = $(top_builddir)/src/libqueso.la
+hessian_eg_CPPFLAGS = $(AM_CPPFLAGS)
+
+dist_custom_tk_DATA =
+dist_custom_tk_DATA += ${hessian_eg_SOURCES}
+dist_custom_tk_DATA += ${hessian_eg_HEADERS}
+
+
 if CODE_COVERAGE_ENABLED
   CLEANFILES = *.gcda *.gcno
 endif
--- a/examples/custom_tk/hessian_eg.C
+++ b/examples/custom_tk/hessian_eg.C
@@ -4,7 +4,7 @@
 // QUESO - a library to support the Quantification of Uncertainty
 // for Estimation, Simulation and Optimization
 //
-// Copyright (C) 2008-2015 The PECOS Development Team
+// Copyright (C) 2008-2017 The PECOS Development Team
 //
 // This library is free software; you can redistribute it and/or
 // modify it under the terms of the Version 2.1 GNU Lesser General
--- a/examples/custom_tk/hessian_eg.C
+++ b/examples/custom_tk/hessian_eg.C
@@ -59,9 +59,6 @@ MyTransitionKernel<V, M>::updateTK()
 
     new_matrix(0, 0) = 1.0;
 
-    // Changing the matrix, so we set the dirty flag so AM can reset
-    this->setCovMatrixIsDirty(true);
-
     // Update all the RVs with new cov matrix (including all the ones used in
     // Delayed Rejection)
     this->updateLawCovMatrix(new_matrix);
@@ -69,6 +66,20 @@ MyTransitionKernel<V, M>::updateTK()
 
   // Just keeping track of a counter so I can trigger a 'dirty' matrix
   m_counter++;
+}
+
+template <class V, class M>
+bool
+MyTransitionKernel<V, M>::covMatrixIsDirty()
+{
+  return m_counter == 15;
+}
+
+template <class V, class M>
+void
+MyTransitionKernel<V, M>::cleanCovMatrix()
+{
+  // Don't need to do anything since we handled it in covMatrixIsDirty
 }
 
 // Explicit instantiation of the template
--- a/examples/custom_tk/hessian_eg.C
+++ b/examples/custom_tk/hessian_eg.C
@@ -39,7 +39,8 @@ MyTransitionKernel<V, M>::MyTransitionKernel(
   const M&                       covMatrix)
   :
   ScaledCovMatrixTKGroup<V, M>(prefix, vectorSpace, scales, covMatrix),
-  m_vectorSpace(vectorSpace)
+  m_vectorSpace(vectorSpace),
+  m_counter(0)
 {
   std::cout << "Hello from MyTransitionKernel!" << std::endl;
 }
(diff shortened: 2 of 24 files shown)


pbauman:
How do they currently interact (if at all)?

dmcdougall:
It really depends on what the user does inside `update_tk`.  For standard random walk, for example, delayed rejection just multiplies the covariance matrix of the Gaussian proposal (whatever it may be) by a scale factor.

For adaptive Metropolis, the covariance matrix is updated with the sample mean of the Markov chain.  This is regardless of whether or not the user modified it in `update_tk`.

dmcdougall:
I think a good first approach would be to, if the user asks for adaptivity, to do it while the user didn't change the proposal covariance matrix inside the `update_tk` method.  This is a little hacky.  Maybe we'd store a copy of the matrix and check it against the matrix after a call to `update_tk` and see if it changed.  Or perhaps a better approach would be to just have a flag somewhere in a "covariance matrix setter" the user has to call that sets the flag.  Then QUESO can just check the flag.

dmcdougall:
Perhaps this is not the right approach at all.  The connection between sampler iteration and transition kernel is being made available to the user here, but having QUESO detect whether or not the user touched the covariance matrix seems too intimate of a connection.  It assumes an awful lot about what the user might have done in the `update_tk` method.  Perhaps the right approach is to have QUESO give the transition kernel the adapted covariance matrix and let the user decide what they want to do with it.

pbauman:
This is starting to sound like the right thing. I'm happy to chat to be a sounding board.

dmcdougall:
Another thing that came to me is that perhaps I should be clear about what motivated this.  A customer wants to modify their proposal covariance matrix as a function of sampler iteration.  They want to do this because they want to be able to change the proposal covariance matrix as the sampler progresses.

In light of this, a good question would be, 'OK, but why does it have to be a *function* of sampler iteration?'  We have all the infrastructure in the `lnValue` parameter list to allow the user to pass higher-order derivative information back to QUESO.  I could make the argument that if QUESO needs Hessian information to update the proposal covariance matrix, then QUESO will ask for it.  If it's the case that our `BaseScalarFunction` interface isn't rich enough, then maybe the right thing to do is update that instead.

Doing this, the user might respond with, "I don't want to eat the cost of constructing the Hessian every iteration."  My response this might be, "Let's implement an input file option telling Hessian-informed algorithms how often they're allowed to query Hessian information."

Sorry if this ticket is a stream of consciousness.  Trying to get my thoughts out so I don't forget them.

roystgnr:
> Perhaps the right approach is to have QUESO give the transition kernel the adapted covariance matrix and let the user decide what they want to do with it.

This sounds right, both from performance and flexibility standpoints.

mseldre:
One comment is that the conversion from QoI Hessian information into the proposal covariance (inverse of the Hessian of the negative log posterior) is not entirely trivial.  There are a few options to consider (full Newton, Gauss-Newton approx of misfit Hessian) and knobs to set (tolerance for eigenvalue truncation when decomposing the Hessian of the prior-preconditioned misfit).

Details of the Dakota implementation for this are in Section 5.2 at 
https://dakota.sandia.gov/sites/default/files/docs/6.5/Theory-6.5.0.pdf

Having said that, we do have the infrastructure to provide both the likelihood and prior density Hessians, if you want to do the calculations on your side.  But I wouldn't advocate this direction at this point.

roystgnr:
After talking to @dcmcdougall more offline, I'm thinking that the right way to do this in the long term is with a strategy pattern interface.  Move the existing proposal covariance adaptivity into strategy functor subclass(es) in QUESO; users could then replace the object of that class with their own strategy functor to get a completely different behavior, or could shim the default object to make more minor tweaks.

dmcdougall:
To summarise the phone conversation between @mseldre, @hovr2pi and myself, we decided not to make them iteration-dependent, but instead let QUESO decide when to ask for the tk and later provide an input file option to let the user tell QUESO how often they'd like QUESO to update it.

dmcdougall:
Ok, I've removed the iteration-dependence and kept the `update_tk` method to allow the user to implement state-dependent updates.  Queso will call `update_tk` every iteration.  Here the idea is that we'll add an input file option to allow the user to specify how often this is called.

I've also added an example.  See the files in `examples/custom_tk/` for what is required from the user.

dmcdougall:
@mseldre Does the provided example look reasonable to you?  The user is required to implement an `update_tk` method in some RV subclass and register it with the factory.

dmcdougall:
So there are some clear downsides with how we're handling the factories right now.  One is that, if you want to update your transition kernel in the `update_tk` method, [you will have to update the proposal covariance matrices for all the transition kernels in the delayed rejection logic](https://github.com/dmcdougall/queso/blob/e0874d3749ec7b576fdaa730aac8e3a818479c18/examples/custom_tk/hessian_eg.C#L232-L242).

If there's not other feedback for this, I think I would like to merge it with the view of improving the interaction between custom transition kernels and delayed rejection.

Any push-back?

briadam:
Based on discussion with @dmcdougall on 20170206 (and of interest to @mseldre ), the design seems sound to me.  The delayed rejection remains well-encapsulated in TK, while the sequence generator retains responsibility for adaptive metropolis accumulation and update, calling update_tk() at a default or user-specified interval.

It seems reasonable that a client specializing a TK, update_tk() in particular would be responsible for appealing to base class functions to ensure that any related TK data like DR stages are updated appropriately.

The interaction of the update_tk interval with things like AM no-adapt period and subsequent AM adapt interval could get a bit confusing and will need documentation.

We also discussed how to handle the AM accumulation (covariance calculation based on chain history), which might be invalidate by update_tk mucking with the proposal covariance, with options including (1) requiring the user to turn it off, (2) adding a m_dirtyCovariance flag to TK or other coupling information, (3) auto-detecting covariance matrix change.  We recognize it's not practical to just assume any call to update_tk() invalidates the AM accumulation history since the default will be to call update_tk every step.  Some variant of (2)/(3) is the most likely candidate, likely for a next development iteration.

briadam:
I discussed this with @mseldre today and overall the proposed changes sound like they'll meet his goals. So go for it!  

From our client perspective, we'd essentially want to reset the AM accumulation history every time update_tk() were called and we provide an updated proposal covariance, so the coupling/detection approaches would probably satisfy that.  (For our case, assuming the AM gets reset at every update_tk interval would suffice, but we realize that's not general.)

What I didn't think of before talking with Mike is that it wasn't clear to me if update_tk() is also the mechanism by which adaptive metropolis updates to the proposal covariance are applied.  I thought @dmcdougall said that update_tk() could be called at some user-specified interval, with AM updates potentially applied to the TK in-between, which would be our preferred behavior when both options are in use.  

Mike and I mused about whether the AM accumulation history might still be useful/valid even in the face of a user-provided proposal covariance via update_tk().

dmcdougall:
Ok, initial AM reset is implemented.

I need to do a little testing, but I thought I'd throw it up here in the off-chance you'd want to look at it.

dmcdougall:
I fixed up a few bugs (related to the interplay of adaptivity interval and when the user dirties the covariance matrix).

It turns out that, since this is pretty buried in the guts of the MCMC sampler, it's pretty hard to write a test to check that adaptivity happens in the right place.  What I'll do is write a regression test to check for a change in future behaviour.

dmcdougall:
@briadam Ready for your feedback.

dmcdougall:
I've made the `covMatrixIsDirty` and `cleanCovMatrix` methods pure virtual, so the user is forced to implement these for their own TK.  All of QUESO's built in TKs implement these to do the right thing, `return false` and do nothing, respectively.

I'm now working on adding the input file option to control how often `updateTK` is called.

dmcdougall:
@briadam Update frequency is done.

Are we good to merge?

briadam:
Yes, I just did a more thorough code review of all the changes in this PR and like how the capability has come together.  Looks good to merge by me!

PR #474, opened 2016-11-03
Require libmesh to be configured with SLEPc

roystgnr:
If libMesh doesn't have SLEPc, then we can't do an eigendecomposition of any operators, so we don't have a basis, so we might as well just give up on infinite dimensional problems entirely.  Configure time is the best time to give up on things entirely, so we'll do detection and report success or failure there.


dmcdougall:
Thanks.  Once travis does its thing I'll merge this.

PR #478, opened 2016-11-17
GPMSA lambda_y hyperprior setting example

roystgnr:
This has always been possible to do, but it's been opaque enough that it made our TODO list, so let's add an example that demonstrates how it's done.

### DIFF
--- a/examples/gp/vector/gpmsa_input.txt
+++ b/examples/gp/vector/gpmsa_input.txt
@@ -48,6 +48,7 @@ ip_mh_am_epsilon                    = 1.e-5
 ip_mh_filteredChain_generate             = 0
 
 # Necessary to avoid sample at NaN domain values?
+ip_mh_algorithm = random_walk
 ip_mh_doLogitTransform = 0
 
 gpmsa_emulator_precision_shape = 5.0
--- a/examples/gp/vector/gpmsa_input.txt
+++ b/examples/gp/vector/gpmsa_input.txt
@@ -48,6 +48,7 @@ ip_mh_am_epsilon                    = 1.e-5
 ip_mh_filteredChain_generate             = 0
 
 # Necessary to avoid sample at NaN domain values?
+ip_mh_tk = random_walk
 ip_mh_algorithm = random_walk
 ip_mh_doLogitTransform = 0
 
--- a/examples/gp/vector/gpmsa_vector.C
+++ b/examples/gp/vector/gpmsa_vector.C
@@ -4,6 +4,7 @@
 #include <queso/StatisticalInverseProblem.h>
 #include <queso/VectorSet.h>
 #include <queso/GPMSA.h>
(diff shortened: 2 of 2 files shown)


roystgnr:
The bad news is that there's been a regression in this example, not from this PR, but since the last time I've run it.  I'm bisecting now.

roystgnr:
And now I'm stumped.  Checking out past commits and cherry-picking 2140331 to fix the compatibility issue, I find that:

483bf1bd6ff85bea2b61bcc8f6cf29828e0dfade, the last commit before #466 was merged, gives me no errors.
a8ba0f8f2f27c13cc0e8146721db206591d8acc9, the last commit within #466, gives me no errors.
18ed273a608a4cfc24b10fba5d3e1a7f9c98055e, the merge commit which just combines the other two, gives me NaNs.

roystgnr:
If I start from 483bf1b, cherry-pick 2140331, and then cherry-pick the rest of the #466 commits in order, then things work after I cherry-pick f92b290 but NaN after I cherry-pick 882115b.  Examining the diff in 	882115b reveals to me no possible reason why this could be happening, but it seems pretty reliable: revert it and things work again, reapply it and things break again.

@dmcdougall, any chance you can take it from there?

dmcdougall:
Looking into it now.

dmcdougall:
For me, this example doesn't run in its current state.  I don't get NaNs, but it does trigger an assertion:

```
Assertion `discrepancy_precision > 0' failed.
discrepancy_precision = 0
0 = 0

../../src/gp/src/GPMSA.C, line 310, compiled Nov 18 2016 at 10:06:30
```

> And now I'm stumped. Checking out past commits and cherry-picking 2140331 to fix the compatibility issue, I find that:

> 483bf1b, the last commit before #466 was merged, gives me no errors.
> a8ba0f8, the last commit within #466, gives me no errors.
> 18ed273, the merge commit which just combines the other two, gives me NaNs

For me:

`483bf1b` works
`a8ba0f8` works
`18ed273` triggers the assertion

dmcdougall:
More confirmation:

If I checkout f92b290 everything's fine
If I then rebase onto 483bf1b (and cherry-pick 2140331) everything's fine
If I then cherry-pick 882115b I trigger the assertion

dmcdougall:
@roystgnr Try adding `ip_mh_tk = random_walk` to the example input file.  Do you have any idea why you aren't triggering the assertion instead of getting NaNs?

roystgnr:
That fixes it for me.

I get the assertion with -O0, and I get NaNs with -O2, despite not using -DNDEBUG in either case.  WTF.

roystgnr:
This should be good to merge once Travis agrees.

IMHO we need some better error handling here, though...

dmcdougall:
> IMHO we need some better error handling here, though...

Totally agree.  Will open a ticket for it.

Issue #14, opened 2013-09-06
Remove  example infoTheoryProblem/ from Queso distribution


kemelli:
This example was created by Gabriel when using library ANN. 
It needs to be tested and documented, so while that is not done, we should remove it from Queso distribution.

dmcdougall:
Fixed in #486.

PR #486, opened 2016-11-28
Clean up and document InfoTheory infrastructure

dmcdougall:
This pull request

1.  Documents the information theoretic methods (to the best of my ability)
2.  Makes the `--enable-ann` configure option work out-of-source
3.  Provides information theoretic template function implementations in an `InfoTheory_impl.h` file to users may use these methods with random variables that aren't just Gaussian (as is the case in the InfoTheory example)

I'm hoping this addresses #14 enough that we can close that ticket.


examples/Makefile.am
```diff
@@ -6,10 +6,13 @@ AM_CPPFLAGS += -I$(top_builddir)/inc
...
 AM_CPPFLAGS += $(GSL_CFLAGS)
 AM_CPPFLAGS += $(GRVY_CFLAGS)
-AM_CPPFLAGS += $(ANN_CFLAGS)
 
 LIBS += $(GSL_LIBS)
 
+if HAVE_ANN
+AM_CPPFLAGS += $(ANN_CFLAGS)
+endif
+
```
roystgnr: Why move this below the GSL_LIBS line?

dmcdougall: No reason other than it's with the other conditional `AM_CPPFLAGS` adjustments.  Want me to move it back up?

src/stats/src/InfoTheory.C
```diff
@@ -22,15 +22,14 @@
 //
 //-----------------------------------------------------------------------el-
 
-#include <queso/InfoTheory.h>
-
 #include <queso/Defines.h>
-#include <gsl/gsl_sf_psi.h> // todo: take specificity of gsl_, i.e., make it general (gsl or boost or etc)
 
 #ifdef QUESO_HAS_ANN
+#include <gsl/gsl_sf_psi.h> // todo: take specificity of gsl_, i.e., make it general (gsl or boost or etc)
```
roystgnr: It's good practice to put third-party dependency headers below own-library headers.  Prevents you from accidentally omitting that dependency in any own-library headers that need it.

src/stats/src/VectorRV.C
```diff
@@ -28,6 +28,7 @@
 
 #include <queso/Defines.h>
 #include <gsl/gsl_sf_psi.h> // todo: take specificity of gsl_, i.e., make it general (gsl or boost or etc)
+#include <queso/InfoTheory.h>
```
roystgnr: This is likely *adding* dependencies, not removing them as the git comment states.  Anything that included VectorRV.h but not any of its subclasses (imagine user code with a container of pointers to the base class) now has a dependency on InfoTheory.h that it didn't have before.

roystgnr: I'm an idiot.  This is VectorRV.C, not .h

roystgnr:
:+1:

Everything looks decent.

dmcdougall:
@roystgnr Cheers.

PR #490, opened 2016-12-02
Fix lognormal normalisation

### LINKED ISSUE #423: Log normal pdf evaluations are incorrect for nonzero `m_normalizationStyle`
While I was implementing #422, I noticed that when `m_normalizationStyle` is nonzero, the pdf evaluation in log space will be incorrect.  Specifically, the term `-std::log(domainVector[i])` isn't in the `returnValue` computation.

dmcdougall:
Fixes #423, and tries to document `m_normalizationStyle`.


dmcdougall:
Woah, I think I branched off of the wrong branch.  Abort, abort!

PR #492, opened 2016-12-02
Remove `typename' for non-dependent types

### LINKED ISSUE #489: GCC 4.4.7 compat: typename in non-templated classes
As follow-up to #488, the following need review:

A number of SharedPtr had typename prefix in non-templated classes.  Of the remaining instances of `typename SharedPtr|ScopedPtr`, I believe the following files also need review, but I'm not building these files for Dakota, so can't edit/test.
```
src/core/src/InfiniteDimensionalMCMCSampler.C
src/core/src/LibMeshOperatorBase.C
src/core/src/LibMeshFunction.C
src/core/src/LibMeshNegativeLaplacianOperator.C
src/core/inc/InfiniteDimensionalMCMCSampler.h
src/core/inc/LibMeshFunction.h
src/core/inc/LibMeshOperatorBase.h
```
These uses appear fine (don't need review) as they are in templated classes
```
src/core/inc/Factory.h
src/gp/inc/GPMSA.h
src/stats/inc/MetropolisHastingsSG.h
src/surrogates/inc/InterpolationSurrogateIOBase.h
```

dmcdougall:
This is fine, but it triggers a compiler error in older versions of gcc.

Fixes #489.

@briadam Look good?

### DIFF
--- a/src/core/inc/InfiniteDimensionalMCMCSampler.h
+++ b/src/core/inc/InfiniteDimensionalMCMCSampler.h
@@ -80,7 +80,7 @@ public:
   unsigned int iteration() const;
 
   //! Returns a pointer to new sampler, with all the moments reset.
-  typename SharedPtr<InfiniteDimensionalMCMCSampler>::Type clone_and_reset() const;
+  SharedPtr<InfiniteDimensionalMCMCSampler>::Type clone_and_reset() const;
 
 private:
   // Current iteration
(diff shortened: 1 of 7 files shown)


briadam:
Looks good to me; thanks!  You probably build LibMesh* files in your testing, but I didn't see anything that actually builds that InfiniteDimensional* stuff.  I was gun shy to remove them all like you did in case some were actually needed, particularly in the .C files.  But if you're happy, I'm happy!

PR #493, opened 2016-12-05
Implement finite difference fallback for `BaseScalarFunction::lnValue`

### LINKED ISSUE #475: Make MALA fall back to finite difference if user does not provide gradient

dmcdougall:
The pull request introduces a bunch of new virtual methods in `BaseScalarFunction` to allow the user to refuse to implement gradient-informed likelihoods, forcing QUESO to fall back to a finite difference approximation instead.  We've traded a compile-time error (not overriding a pure virtual method in a class that will be instantiated) with a run-time error (throw an exception when the user hasn't implemented any of the old or new `lnValue` methods), but I believe it's worth it given the added functionality.

Finite difference fallbacks for Hessian information are not implemented.  If QUESO asks for Hessian information and the user hasn't implemented it, an exception is thrown.  We suspect the user would never tolerate the a) inaccuracy; and b) expense of a first order finite difference approximation to a Hessian operator.

Test has been added.

I'm taking input as to how the finite difference step size should be set.  I'm currently leaning towards a member variable that is set programatically, but I'll also accept that another viable option is to have it set via an input file option.

This pull request fixes #475.

### DIFF
--- a/src/basic/src/ScalarFunction.C
+++ b/src/basic/src/ScalarFunction.C
@@ -52,6 +52,56 @@ const VectorSet<V, M> & BaseScalarFunction<V, M>::domainSet() const
   return m_domainSet;
 }
 
+template <class V, class M>
+double
+BaseScalarFunction<V, M>::lnValue(const V & domainVector,
+                                  const V * domainDirection,
+                                  V * gradVector,
+                                  M * hessianMatrix,
+                                  V * hessianEffect) const
+{
+  std::string msg;
+
+  msg += "Implementation of all lnValue methods is missing.  Please implement";
+  msg += "at least lnValue(const V &).";
+
+  queso_error_msg(msg);
+}
+
+template <class V, class M>
+double
+BaseScalarFunction<V, M>::lnValue(const V & domainVector) const
+{
+  return this->lnValue(domainVector, NULL, NULL, NULL, NULL);
+}
+
+template <class V, class M>
+double
+BaseScalarFunction<V, M>::lnValue(const V & domainVector, V & gradVector) const
+{
+  // Read this from an input file, or make it a settable member variable.
+  double h = 1e-6;
+
+  double value = this->lnValue(domainVector);
+
+  // Create perturbed version of domainVector to use in finite difference
+  V perturbedVector(domainVector);
+
+  // Fill up gradVector with a finite difference approximation
+  for (unsigned int i = 0; i < domainVector.sizeLocal(); ++i) {
+    // Store the old value of the perturbed element so we can undo it later
+    double tmp = perturbedVector[i];
+
+    perturbedVector[i] += h;
+    gradVector[i] = (this->lnValue(perturbedVector) - value) / h;
+
+    // Restore the old value of the perturbedVector element
+    perturbedVector[i] = tmp;
+  }
+
+  return value;
+}
+
 }  // End namespace QUESO
 
 template class QUESO::BaseScalarFunction<QUESO::GslVector, QUESO::GslMatrix>;
--- a/src/basic/src/ScalarFunction.C
+++ b/src/basic/src/ScalarFunction.C
@@ -24,11 +24,14 @@
 
 #include <queso/ScalarFunction.h>
 #include <queso/VectorSet.h>
+#include <queso/VectorSpace.h>
 #include <queso/GslVector.h>
 #include <queso/GslMatrix.h>
 #include <queso/BoostInputOptionsParser.h>
 
-#define QUESO_BASESCALARFN_FD_STEPSIZE_ODV 1e-6
+#include <cstdlib>
+
+#define QUESO_BASESCALARFN_FD_STEPSIZE_ODV "1e-6"
 
 namespace QUESO {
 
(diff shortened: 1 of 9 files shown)


pbauman:
> I'm taking input as to how the finite difference step size should be set. I'm currently leaning towards a member variable that is set programatically, but I'll also accept that another viable option is to have it set via an input file option.

Why not both?

dmcdougall:
> Why not both?

I'll do both.  Hold off on merging this until I get that sorted out.

dmcdougall:
Done.

src/stats/inc/BayesianJointPdf.h
```diff
@@ -97,6 +97,11 @@ class BayesianJointPdf : public BaseJointPdf<V,M> {
 
   //@}
 
+  //! @name Using declarations
+  //@{
+  using BaseScalarFunction<V, M>::lnValue;
+  //@}
```
roystgnr: Damn, I didn't think of this.  Will user classes be broken by the same problem?

src/basic/src/ScalarFunction.C
```diff
@@ -35,8 +38,27 @@ BaseScalarFunction<V, M>::BaseScalarFunction(const char * prefix,
...
+#ifndef DISABLE_BOOST_PROGRAM_OPTIONS
+  m_parser->registerOption<double>(m_prefix + "fdStepSize",
+                                   QUESO_BASESCALARFN_FD_STEPSIZE_ODV,
+                                   "step size for finite difference");
+  m_parser->scanInputFile();
+  m_parser->getOption<double>(m_prefix + "fdStepSize", m_fdStepSize);
+#else
+  m_fdStepSize = m_env.input()(m_prefix + "fdStepSize",
+                               QUESO_BASESCALARFN_FD_STEPSIZE_ODV);
+#endif
```
roystgnr: This is a good start, but it isn't always going to be fine-grained enough, at least not for users who aren't normalizing all their parameters.  When trying to do UQ with both a reaction rate preexponential coefficient of ~1e11 and a nitridation rate of ~0.3, there's no way to come up with a single perturbation distance that works well for both.

Take a look at the numerical_jacobian_h* stuff in https://github.com/libMesh/libmesh/blob/master/include/systems/fem_system.h

We let users override a gross numerical_jacobian_h, but optionally then they can also (or instead) individually override numerical_jacobian_h_for_var on one (or two or all) of the system's variables.  This wasn't coded out of forethought, either, this was necessary to get proper quasi-Newton convergence rates on some nasty un-normalized physics.

src/basic/src/ScalarFunction.C
```diff
@@ -92,8 +111,8 @@ BaseScalarFunction<V, M>::lnValue(const V & domainVector, V & gradVector) const
     // Store the old value of the perturbed element so we can undo it later
     double tmp = perturbedVector[i];
 
-    perturbedVector[i] += h;
-    gradVector[i] = (this->lnValue(perturbedVector) - value) / h;
+    perturbedVector[i] += m_fdStepSize;
+    gradVector[i] = (this->lnValue(perturbedVector) - value) / m_fdStepSize;
```
roystgnr: You can get a little more flexibility in most cases by doing central rather than one-sided differences.  2N instead of N evaluations, but epsilon^2 instead of epsilon error.

On the other hand, there are a lot of parameters where the domain is "non-negative numbers" and where the evaluation point is "just barely non-negative", and there is something to be said for only doing perturbations in the positive direction.

src/stats/inc/BayesianJointPdf.h
dmcdougall: Yes and no.  Though, technically yes.  Here's the deal:

If the user didn't call `lnValue` in their application code, then it's fine.  QUESO interacts with a `BaseScalarFunction` type for run-time polymorphism, and in that class everything is declared so no `using` declaration is needed.

If the user calls `lnValue` (for whatever reason) in their application code, then they'll need to add a `using` declaration to prevent the derived class from hiding all the other `lnValue` names declared in the base class.

Typically the user doesn't call `lnValue`.  QUESO does.

Will this fix the breakage?
1.  Add a super class to `BaseScalarFunction`, call it `A`;
2.  Move all the declarations of `lnValue` from `BaseScalarFunction` to `A`;
3.  Add a `using` declaration to `BaseScalarFunction` to reveal all the `lnValue` names;
4.  Hope.

I can try this locally to see if it works.

dmcdougall: That idea does not fix the breakage.

roystgnr:
If the breakage only occurs for user code calling itself then I wouldn't worry about it.  That's not an expected use case.

src/basic/src/ScalarFunction.C
dmcdougall: I could make it a vector instead.  That way if the user sets it with a scalar value, it just sets every element.  Or are you thinking something even more flexible where the user can set every element with a scalar value and then override ones for certain components individually?

dmcdougall: Here's a compromise: make the finite difference step size a vector.  If the user sets a scalar in the input file then it'll set every component of the finite difference step size vector, but I won't bother adding logic to set individual components in the input file (it's a little more work).  Instead, I'll add some member functions that let you set individual component of the finite difference step size vector.

roystgnr: That should work.

dmcdougall: Done in 919994e.

roystgnr:
:+1:

pbauman:
Sorry I'm late to the party here, but just to clarify: is the intent to always have the user provide the Hessian-effect and not provide a function-to-be-overridden that would allow the user to populate a Hessian matrix? It doesn't really affect me one way or the other at this point, just more of trying to understand what the plans are so that I may keep them in mind.

Also, what are the prevailing thoughts on `actualValue`? Is it used in the library anywhere?

Issue #494, opened 2016-12-09
Consolidate configured headers


briadam:
Suggested enhancement: What do you think about consolidating config_queso.h and queso.h, or at least reworking them so they each have include guards and don't redefine any duplicate pre-processor names?  

There isn't a particular bug right now, just potential for confusion, e.g., when the static configure-time generated BUILD_DATE in queso.h doesn't get overridden by the compile time one in config_queso.h.

dmcdougall:
If I'm honest, I'm struggling to think of a reason why we needed both to begin with.

briadam:
Okay I can do a review of which defines are actually used and propose something.  A few of the things in queso.h might not really be "configured," but probably not too much of an abuse to put in config_queso.h.

dmcdougall:
I propose nuking the dupes in `queso.h` and adding `#include <queso_config.h>` in `queso.h` assuming `config_queso.h` is installed.

Just in case anyone is actually using `queso.h`.

pbauman:
I would suggest throwing in a `#warning` in the deprecated header for those that are actually using the header.

briadam:
@dmcdougall If it fits with your workflow, could the changes on libqueso:hotfix-0.56.1 be merged into dev?  I presume I should make changes to queso.h on dev? It would help to have all the typename stuff there...

dmcdougall:
I could certainly argue that's a bug fix.  Sounds good to me.

briadam:
Sorry I should have asked a question instead of stating something.  Which branch should my pull request be based on to resolve this issue #494?

dmcdougall:
Branch off of `hotfix-0.56.1`.

`hotfix-0.56.1` will be merged to both `master` and `dev`.  I think once this is resolved we can do a `v0.56.1` release.  Does that sound reasonable?

PR #495, opened 2016-12-13
Consolidate configured headers #494

briadam:
Proposal to address #494.  Needs careful scrutiny as my autotools skills are super rusty.  I did verify that make install, and make dist all still work, but know my build environment isn't synced with yours, so will rely on Travis for builds.

One thing I wasn't sure of is if you want the TPL-specific AC_DEFINES in their .m4 macros vs in config_environment.m4; they seemed to kind of straddle TPL information vs. QUESO-specific defines.

Commit log:

Make queso.h a static, deprecated header that depends on a single
configured config_queso.h.  For now, queso.h is still packaged and
installed.  The various definitions in queso.h.in are dealt with as
follows:

```
QUESO_MAJOR_VERSION: def in config_queso.h in terms of GENERIC_MAJOR_VERSION
QUESO_MINOR_VERSION: def in config_queso.h in terms of GENERIC_MINOR_VERSION
QUESO_MICRO_VERSION: def in config_queso.h in terms of GENERIC_MICRO_VERSION

QUESO_BUILD_USER: already defined in config_queso.h
QUESO_BUILD_ARCH: already defined in config_queso.h
QUESO_BUILD_HOST: already defined in config_queso.h
QUESO_BUILD_DATE: already defined in config_queso.h
QUESO_BUILD_VERSION: already defined in config_queso.h

QUESO_LIB_VERSION: deprecated; use QUESO_VERSION instead
QUESO_LIB_RELEASE: deprecated; use QUESO_BUILD_DEVSTATUS instead

QUESO_CXX: def in config_queso.h in terms of CXX
QUESO_CXXFLAGS: def in config_queso.h in terms of CXXFLAGS

QUESO_TRILINOS_DIR: def in config_queso.h in terms of TRILINOS_HOME
QUESO_GSL_DIR: def in config_queso.h in terms of GSL_LIBS
QUESO_GRVY_DIR: def in config_queso.h in terms of GRVY_PREFIX
QUESO_GLPK_DIR: def in config_queso.h in terms of GLPK_PREFIX
QUESO_HDF5_DIR: def in config_queso.h in terms of HDF5_PREFIX
```


src/Makefile.am
```diff
@@ -512,7 +511,6 @@ libqueso_include_HEADERS += contrib/getpot/getpot.h
 
 nodist_libqueso_include_HEADERS =
 nodist_libqueso_include_HEADERS += $(top_builddir)/config_queso.h
-nodist_libqueso_include_HEADERS += $(top_builddir)/src/core/inc/queso.h
```
dmcdougall: I'm a little confused.  Not by what you did, but by what we were doing.   It seems that before this change, neither `queso.h` nor `config_queso.h` were shipped in our tarballs.  If that's the case (and I will check that), then it means users could never `#include` these files anyway.

dmcdougall: (Unless, of course, they built and installed `QUESO` from the repository source tree)

dmcdougall: Oh no, I misunderstood.  Of course we don't ship `config_queso.h` in our tarballs because it's generated by the build system when the user does `./configure && make install`.

As for `queso.h`, I can verify it *is* in the latest (v0.56.0) tarball, so it's not clear to me why the `nodist_` prefix had no effect.

configure.ac
```diff
@@ -26,6 +26,12 @@ GENERIC_MAJOR_VERSION=$AX_MAJOR_VERSION
 GENERIC_MINOR_VERSION=$AX_MINOR_VERSION
 GENERIC_MICRO_VERSION=$AX_POINT_VERSION
 
+# Defines needed by Environment.C (formerly in queso.h)
+# The values are not quoted strings per queso.h history (Environment.C uses for arithmetic)
+AC_DEFINE_UNQUOTED([GENERIC_MAJOR_VERSION], ${GENERIC_MAJOR_VERSION}, "Major version")
+AC_DEFINE_UNQUOTED([GENERIC_MINOR_VERSION], ${GENERIC_MINOR_VERSION}, "Minor version")
+AC_DEFINE_UNQUOTED([GENERIC_MICRO_VERSION], ${GENERIC_MICRO_VERSION}, "Micro version")
```
dmcdougall: Correct me if I'm wrong, but I believe the first argument to these three macro calls should not have the `GENERIC_` prefix.  Otherwise users that relied on `QUESO_MAJOR_VERSION` from `queso.h` will need to change their code to use `QUESO_GENERIC_MAJOR_VERSION`, for example.

briadam: Yes, that looks correct.  I'm not sure how I missed that.  I can push that change.

src/Makefile.am
briadam: Yeah not sure I get that either.  Autotools is ancient history for me, but I'm guessing that libqueso_include_HEADERS=.../queso.h causes it to get packaged while nodist_libqueso_include_HEADERS says to install, but not package it.  I would guess these two options augment each other, rather than override each other, so the net effect is queso.h ends up both packaged in distributions and installed.  In constrast config_queso.h is only called out as nodist, so it would get installed, but not packaged.

PR #498, opened 2016-12-16
Tidy up `*Environment` memory management

dmcdougall:
This pull request wraps up all internal raw pointers in `ScopedPtr` objects.  Better memory management and better exception safety for free.

I had to compromise on one thing, which is the options object.  Instead of just taking the pointer the user passes and throwing it in a `ScopedPtr` we copy the user's object and throw the copy in a `ScopedPtr`.  Options objects aren't too big and are only copied once during the construction of `BaseEnvironment`.

Feedback welcome.

### DIFF
--- a/src/core/src/Environment.C
+++ b/src/core/src/Environment.C
@@ -150,8 +150,8 @@ BaseEnvironment::BaseEnvironment(
   m_optionsInputFileName       (""),
   m_optionsInputFileAccessState(true),
 #ifndef DISABLE_BOOST_PROGRAM_OPTIONS
-  m_allOptionsDesc             (NULL),
-  m_allOptionsMap              (NULL),
+  m_allOptionsDesc             (),
+  m_allOptionsMap              (),
 #endif  // DISABLE_BOOST_PROGRAM_OPTIONS
   m_input                      (new GetPot),
   m_subComm                    (),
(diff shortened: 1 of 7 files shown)


dmcdougall:
Before:

```
definitely lost: 880 bytes in 3 blocks
indirectly lost: 33,760 bytes in 511 blocks
```

After:

```
definitely lost: 128 bytes in 2 blocks
indirectly lost: 27,411 bytes in 412 blocks
```

src/core/src/Environment.C
```diff
@@ -1210,7 +1209,7 @@ FullEnvironment::construct (RawType_MPI_Comm inputComm,
   int mpiRC = MPI_Comm_rank(inputComm,&m_worldRank);
   queso_require_equal_to_msg(mpiRC, MPI_SUCCESS, "failed to get world fullRank()");
 
-  m_fullComm = new MpiComm(*this,inputComm);
+  m_fullComm.reset(new MpiComm(*this,inputComm));
```
roystgnr: No corresponding delete?  This was a straight-up memory leak before?

roystgnr:
:+1:

src/core/src/Environment.C
dmcdougall: It would appear so.

PR #499, opened 2016-12-17
Tidy up `StatisticalInverseProblem` memory management

dmcdougall:


src/stats/src/StatisticalInverseProblem.C
```diff
@@ -141,14 +136,14 @@ StatisticalInverseProblem<P_V,P_M>::StatisticalInverseProblem(
...
-    // m_optionsObj is a pointer to const
-    m_optionsObj = tempOptions;
-
-    m_userDidNotProvideOptions = true;
+    m_optionsObj.reset(new SipOptionsValues(&m_env, prefix));
   }
+  else {
+    // If the user passed in a legitimate options object, then copy it into a
+    // ScopedPtr
+    m_optionsObj.reset(new SipOptionsValues(*alternativeOptionsValues));
```
roystgnr: I agree that this is the right way to do things in this case, but we should document that we're keeping a copy rather than a reference.  Users are going to ask "Can I deconstruct my options object or does it need to outlive the SIP", and users are going to ask "Can I change my options object after the SIP is constructed", and there's no good way to indicate "Yes and no respectively" in code rather than docs.

src/stats/src/StatisticalInverseProblem.C
```diff
@@ -173,7 +173,6 @@ StatisticalInverseProblem<P_V,P_M>::~StatisticalInverseProblem()
   }
   if (m_logLikelihoodValues) {
     m_logLikelihoodValues->clear();
```
roystgnr: This (like its' predecessor) is weird.  Does clear() have side-effects which do not also get invoked via the destructor?  If not then why are we calling it on an about-to-be-destroyed object?  If so then whose mad idea was that??

src/stats/src/StatisticalInverseProblem.C
```diff
@@ -176,7 +176,6 @@ StatisticalInverseProblem<P_V,P_M>::~StatisticalInverseProblem()
   }
   if (m_logTargetValues) {
     m_logTargetValues->clear();
```
roystgnr: Same question, different object.

roystgnr:
:+1:

PR #500, opened 2016-12-19
Use ScopedPtr for members of MetropolisHastingsSG

dmcdougall:
Instead of raw pointers.


src/stats/src/MetropolisHastingsSG.C
```diff
@@ -337,7 +334,7 @@ MetropolisHastingsSG<P_V,P_M>::MetropolisHastingsSG(
   // as a member (m_oldOptions) because otherwise m_ov will die when the
   // MetropolisHastingsSGOptions instance dies.
   m_oldOptions = new MetropolisHastingsSGOptions(mlOptions);
-  m_optionsObj = &(m_oldOptions->m_ov);
+  m_optionsObj.reset(new MhOptionsValues(m_oldOptions->m_ov));
```
roystgnr: Is there a similar documentation issue here?

PR #502, opened 2016-12-19
Use nonrecursive make for the gsl_tests directory

dmcdougall:
Does what it says on the tin.  `make -j12 distcheck` passes locally for me.


dmcdougall:
Trying to transition all our tests to nonrecursive make so we get a cleaner `make check` test summary and parallel test builds are more effective.  Yes, it's just a few tests, but I like everything neat and tidy.  This PR only addresses the `gsl_tests` directory.

roystgnr:
Using non-recursive make is also faster (much faster with -j), more flexible when you want to override the TESTS variable, etc.

In theory the best way to work is with recursive automake files (using include directives so you get modularity) to generate non-recursive Makefiles, but I've never tried to set that up properly in practice.

roystgnr:
:+1:

configure.ac
```diff
@@ -269,7 +269,6 @@ AC_CONFIG_FILES([
   test/t02_sip_sfp/Makefile
   test/t03_sequence/Makefile
   test/t04_bimodal/Makefile
-  test/gsl_tests/input
```
roystgnr: What's up with "Conflicts: configure.ac" in the commit message?

dmcdougall:
> In theory the best way to work is with recursive automake files (using include directives so you get modularity) to generate non-recursive Makefiles, but I've never tried to set that up properly in practice.

This makes total sense, and it never occurred to me.  Thanks.

PR #507, opened 2017-01-06
Supported ConcatenatedRV without realizers in all sub-RVs

### LINKED ISSUE #506: Make ConcatenatedVectorRV implementation requirements mirror those of VectorRV
The constructors for ConcatenatedVectorRV access members of the component VectorRVs that may not be implemented, resulting in runtime errors.  Relevant code is at ConcatenatedVectorRV.C:99 and following.

Suggest having the non-NULL members in ConcatentedVectorRV mirror those of a VectorRV itself.  Or if realizer is required, enforcing that.

Use case: In Dakota I construct my own QuesoVectorRV that has an image set and PDF, but no realizer.  When passing this as the prior to GPMSA it rightly gets concatenated with the hyper parameter priors, but the ConcatenatedVectorRV constructor fails at runtime as noted above.

(Per our discussion @dmcdougall @roystgnr)

roystgnr:
This should fix #506, but I don't know if we have any BaseVectorRV subclasses without realizers in the library itself to test with; let's wait to merge until @briadam tries it out on his case.

### DIFF
--- a/src/stats/inc/VectorRV.h
+++ b/src/stats/inc/VectorRV.h
@@ -76,6 +76,9 @@ public:
   //! Posterior Density Function of the vector RV; access to private attribute m_pdf.
   const   BaseJointPdf      <V,M>& pdf       () const;
 
+  //! Returns true iff this RV has the ability to produce realizations (samples)
+          bool                  has_realizer () const;
+
   //! Finds a realization (sample) of the PDF of this vector RV; access to private attribute m_realizer.
   const   BaseVectorRealizer<V,M>& realizer  () const;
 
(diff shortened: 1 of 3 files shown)


roystgnr:
I can't (and apparently clang can't) replicate that failure.  @dmcdougall  can you?

dmcdougall:
Looks good, @roystgnr.

briadam:
The code changes make sense to me and they work for my concatenated test case where my VectorRV doesn't have a realizer, but the hyper-parameter VectorRV does.  Thanks @roystgnr

PR #508, opened 2017-01-06
Fix symbol-stripping via static linking

dmcdougall:
This fixes a discrepancy in run-time behaviour when linking statically with `libqueso.a` versus linking dynamically with `libqueso.so`.

Before this change, we statically initialise some factories to register various transition kernels and algorithms.  The instantiation occurs before `main` starts executing and so, by the time `main` enters, the `static` factory map contains all the necessary keys and values for all the transition kernels and algorithms.  The shared object file `libqueso.so` contains these symbols, as does `libqueso.a`.

When a user links an application statically against `libqueso.a`, the linker (`ld` in this case) thinks, "Oh, you're not using any of these global variables, despite the fact that your code relies on the side-effects from creating these symbols."  The linker then proceeds to strip them, and the `static` factory map never gets populated.

This pull request does away with static initialisation of those global symbols and instead QUESO triggers their instantiation explicitly before constructing the transition kernel and algorithm.  This is done by throwing all those instantiations into the constructor for `TKFactoryInitializer` in the case of transition kernels and `AlgorithmFactoryInitializer` for algorithms.  These objects should be user-extendable to allow users to trigger instantiation of custom kernels so that they are registered with the factory before QUESO needs them.

@briadam Mind giving this PR a test before merging?

I'm targetting `hotfix-0.56.2` here with the goal to roll this as a bugfix release.  I'm open to push-back if we'd prefer to have this in the next major release instead.


dmcdougall:
Just as a note to others.  These resources helped me figure out what was going on:

http://stackoverflow.com/questions/9459980/c-global-variable-not-initialized-when-linked-through-static-libraries-but-ok

http://www.lysium.de/blog/index.php?/archives/222-Lost-static-objects-in-static-libraries-with-GNU-linker-ld.html

http://stackoverflow.com/questions/5202142/static-variable-initialization-over-a-library

roystgnr:
:+1:

pbauman:
Yikes, I didn't realize this would happen with static linking. We actually need to fix this in grins too. Thanks for the links and explanation.

briadam:
@pbauman We've struggled with related problems in Sandia's Acro and Utilib software with newer Intel compilers, e.g., https://software.sandia.gov/trac/utilib/ticket/4104.  Thankfully the QUESO situation is a little cleaner for @dmcdougall so his patch is a a lot cleaner.

briadam:
@dmcdougall I confirmed that the random walk TK doesn't get registered with the factory in my debug build prior to your commits.  When I include the commits in #508, the debug build works as expected.  Thanks for the quick fix!

roystgnr:
Damn, @pbauman, sorry I didn't bring this to your attention.  I decided the problem wouldn't much affect libMesh (where we only create a Partitioner factory and then we don't even use that internally) but I completely forgot that GRINS is heavily factoried-up now.

Hm... maybe I should double-check libMesh for non-Factory-based instances of this problem?  Do we use global constructors to instantiate the string_to_enum methods' data?

roystgnr:
No, we initialize on first call.  That's the only other suspicious code I can think of off the top of my head... but I wonder if there's any way we can trick nm or objdump into listing what objects are in a shared library expecting to be globally constructed?

dmcdougall:
@briadam Do you want a hotfix release with this in?  Or do you just want to wait until the next QUESO release?

briadam:
Don't have a strong preference.  I won't be able to get back to GPMSA integration until February 2 and at that time it would help us to have it either in a 0.56.x or 0.57.  It seemed there were a few things more hotfix-flavored, but also a few more major changes.  If they're not easy to separate out, waiting for a release is fine.

dmcdougall:
Ok, in that case I'll merge `hotfix-0.56.2` into `dev` and we'll just wait for the next release.

dmcdougall:
Merged in 9f76ec1f2ea49220bd9deece728d3a45401643de.

PR #515, opened 2017-01-13
Add support for `std::random` random number generation

dmcdougall:
Adds support for random variate generation using `std::random`.  To use it, you can set
```
env_rngType = cxx11
```
in the QUESO input file.

1.  You need to have compiled QUESO with C++11 support to use this functionality;
2.  If you haven't done 1., you will get a descriptive run-time error explaining that you need to do 1.;
3.  User code will not break because the default rng type is still `gsl`;
4.  There are regression test solutions that were generated with the GSL rng.  Switching to any other rng will obviously break these, so I've explicitly set these tests to always use `gsl` for the rng until we generate new regression test solutions.


roystgnr:
It's std::mt19937...

How crazy would it be for us to switch RngGsl to use gsl_rng_mt19937 and use the same (new) regression standards for both?

src/core/inc/RngCXX11.h
```diff
@@ -0,0 +1,124 @@
...
+   * Support (domain): [0, infinity).
+   */
+  double gammaSample(double a, double b) const;
+
+private:
+  //! Default Constructor: it should not be used.
+  RngCXX11();
+
+  //! The internal random number generator.  Mersenne Twister generator.
+  mutable std::mt19937 m_rng;
```
roystgnr: We really need to go through at some point and figure out what's thread-safe in QUESO and what isn't.

roystgnr:
Or better, we could leave "gsl" as the env_rngType that means ranldx2, but then add "gsl_mt" or other options that use the different internal GSL RNGs.

roystgnr:
:+1:

dmcdougall:
> How crazy would it be for us to switch RngGsl to use gsl_rng_mt19937 and use the same (new) regression standards for both?

I wasn't explicit about it, but that's what I meant.

> Or better, we could leave "gsl" as the env_rngType that means ranldx2, but then add "gsl_mt" or other options that use the different internal GSL RNGs.

And keep multiple sets of regressions solutions around?  I'm ok with that, it'll just require a bit more book-keeping.

pbauman:
> Or better, we could leave "gsl" as the env_rngType that means ranldx2, but then add "gsl_mt" or other options that use the different internal GSL RNGs.

+1 for more fine-grained control from the input for RNG. IIRC, you can set the RNG algorithm for GSL from a Bash environment option.

PR #516, opened 2017-01-13
GPMSAOptions refactor

### LINKED ISSUE #505: Default ctor for GPMSAOptions
@roystgnr I tried to setup a GPMSA problem without an input file and ran into the issue hinted at by @dmcdougall in GPMSA.C:479.  It looks like the intention is to allow options structure, input file, or none (to get default options).  The code is correctly protecting the case of neither (none) because the GPMSAOptions() constructor taking an input prefix rightly requires an input file (else it dies).

I propose that GPMSAOptions() support a default constructor, paralleling SipOptionsValues().  Then I can default construct the options and only override those I need to through C++.

Note: in my longer-term plan to allow Dakota power users to provide a QUESO input file to set advanced options not exposed to Dakota, I'll need the current functionality, but for testing, it would be helpful to not need the input file.

roystgnr:
We can now default-construct a GPMSAOptions object (which then gets default parameter values), or parse a new input file with such an object (overriding any values found in the input file but leaving alone any values not found).

This should resolve #505

### DIFF
--- a/src/gp/inc/GPMSAOptions.h
+++ b/src/gp/inc/GPMSAOptions.h
@@ -25,6 +25,7 @@
 
 #ifndef DISABLE_BOOST_PROGRAM_OPTIONS
 #include <queso/BoostInputOptionsParser.h>
+#include <queso/ScopedPtr.h>
 #endif  // DISABLE_BOOST_PROGRAM_OPTIONS
 
 #ifndef UQ_GPMSA_OPTIONS_H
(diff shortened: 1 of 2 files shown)


briadam:
I really like this fine-grained control!  This will let me have greater control over the precedence of user options from the Dakota side, since now I have the option of parse, then C++ options set, or vice-versa.  Would be great to support this in other *Options classes, but not a high priority by me.  If it gets urgent for me, I can contribute it...

briadam:
I trial integrated this into Dakota and it looks good to me.  Builds/runs anyway.  Now I have to start implementing all of the options @brianw525 requested...

PR #518, opened 2017-01-14
Switch GPMSA code to new SharedPtr APIs

roystgnr:
There will be a bit more API breakage later, so we might as well improve the memory handling while we're at it.

This branches off of #516, so includes the first two commits from that and shouldn't be merged first.

### DIFF
--- a/examples/gp/pseudovector/gpmsa_pseudovector.C
+++ b/examples/gp/pseudovector/gpmsa_pseudovector.C
@@ -9,11 +9,11 @@
 
 // Read in data files.  Return std deviations for each output.
 std::vector<double>
-readData(const std::vector<QUESO::GslVector *> & simulationScenarios,
-         const std::vector<QUESO::GslVector *> & simulationParameters,
-         const std::vector<QUESO::GslVector *> & simulationOutputs,
-         const std::vector<QUESO::GslVector *> & experimentScenarios,
-         const std::vector<QUESO::GslVector *> & experimentOutputs) {
+readData(const std::vector<QUESO::SharedPtr<QUESO::GslVector>::Type> & simulationScenarios,
+         const std::vector<QUESO::SharedPtr<QUESO::GslVector>::Type> & simulationParameters,
+         const std::vector<QUESO::SharedPtr<QUESO::GslVector>::Type> & simulationOutputs,
+         const std::vector<QUESO::SharedPtr<QUESO::GslVector>::Type> & experimentScenarios,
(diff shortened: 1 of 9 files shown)


dmcdougall:
I'll label `do not merge` so I don't forget to do #516 first.

briadam:
This looks like a great improvement to me @roystgnr .  Only remaining pointer is to the options object, but that's consistent with the other APIs for options, so the client will have to manage the memory for those objects.  Reasonable to me.

dmcdougall:
#516 has been merged.

dmcdougall:
Commit af662a2 breaks backwards compatibility.  How should we handle this?  Since we're not deprecating-then-removing we could release a `v1.0` to follow semantic versioning best practices.  If we do that, there's a ton of other deprecated stuff I'd like to purge (not in this pull request) so we can concentrate all the breaking stuff together in a single release rather than breaking backward compatibility multiple times.

Another option is to add the new methods that take `vector<SharedPtr<V>::Type >`s, deprecate the one that take `vector<V *>` (instead of removing them) and rely on function overloading to not break users' code for the time being.

Thoughts?

briadam:
You're right to generally worry about backward compatibility and have project policies for this kind of API change (I need to manage this better for my projects).  While I realize you don't have a solid way to know for sure, do you think anyone was using the GPMSA code prior to this proposed API change?  I might have misunderstood, but my perception was that it wasn't a complete/reliable implementation and that its use was discouraged.  Anyway, just food for thought; it doesn't give us (Dakota) any heartache to break backward compatibility in this particular interface.

dmcdougall:
I highly doubt anybody was using it.  In fact, I suspect the only instantiation of `GPMSAFactory` outside of Dakota is in our `examples/gp` directory.  We could compromise; break GPMSA backwards compatibility but also make a note of it in our changelog.  I'd be willing to make a bet that as a result of this I'll be able to nothing but roaring silence on our `queso-users` mailing list.

briadam:
And FWIW, code in Dakota that uses it hasn't even been committed to our devel branch.  It's still on a feature branch.  The only interface we had in production Dakota was to Ernesto's code and it was hard-wired to solve a particular CASL application problem, not a general calibration problem.

dmcdougall:
Ah, ok.  Good to know.

@briadam You want to push the green button?

briadam:
With great power comes great responsibility.  I am new enough to your project that I'm not sure everything that should be checked before merging, but this one seems straightforward enough, so here goes...

PR #519, opened 2017-01-16
Add CppUnit-based testing, move some GSL tests to CppUnit, add Legendre1DQuadrature tests

pbauman:
I'm going to be refactoring/extending the QUESO quadrature capabilities. As such, I wanted to have some unit tests in place.  To that end, I went ahead and added (optional) CppUnit-based testing. and 

Build system was updated to check for CppUnit; if QUESO doesn't detect/use it, then the `unit_driver` test is reported as `SKIP`, otherwise, all the CppUnit tests are run under `unit_driver`. I moved my ancient GSL tests (those that were in test/gsl_tests) to CppUnit-based tests and removed the old ones. Also added tests for the 1D Gauss-Legendre quadrature. Serves as a good starting point for adding additional 1D quadrature tests (and testing internal refactoring that's coming down the line). Also provides expansion to more CppUnit-based testing for QUESO as a whole.

I like CppUnit and this style of testing (because it makes it super easy to add unit tests), so I just went ahead and did it. If folks don't like it, I can drop this.


m4/common/cppunit.m4
```diff
@@ -0,0 +1,98 @@
...
+  AC_PATH_PROG(CPPUNIT_CONFIG, cppunit-config, no)
+  cppunit_version_min=$1
+
+  AC_MSG_CHECKING(for Cppunit - version >= $cppunit_version_min)
+  no_cppunit=""
+  if test "$CPPUNIT_CONFIG" = "no" ; then
+    AC_MSG_RESULT(no)
+    no_cppunit=yes
+  else
+    CPPUNIT_CPPFLAGS=`$CPPUNIT_CONFIG --cflags`
```
dmcdougall: This may be nitpicking, but does this conflate compiler flags and preprocessor flags?

pbauman: It does conflate it, but it's CppUnit's conflation and not mine:

<pre>
Usage: cppunit-config [OPTION] ...

Generic options
  --version	output CppUnit version information.
  --help	display this help and exit.

Compilation support options
  --cflags	print pre-processor and compiler flags
  --libs	print library linking information

Install directories CppUnit was configured to
  --prefix[=DIR]      
  --exec-prefix[=DIR]
</pre>

We're sniffing `cppunit-config` to get the `CPPFLAGS` we need.

test/Makefile.am
```diff
@@ -236,11 +230,6 @@ BimodalExample_gsl_CPPFLAGS = -I$(srcdir)/t04_bimodal $(AM_CPPFLAGS)
 
 # Files to freedom stamp
 srcstamp =
-srcstamp += $(test_gsl_get_set_row_column_SOURCES)
-srcstamp += $(test_gsl_get_min_max_vec_SOURCES)
-srcstamp += $(test_gsl_multiple_rhs_matrix_solve_SOURCES)
-srcstamp += $(test_gsl_power_method_SOURCES)
-srcstamp += $(test_gsl_inverse_power_method_SOURCES)
```
pbauman: Actually, how about I just remove `srcstamp` altogether? The file list for stamping was replaced by [foo in configure.ac](https://github.com/libqueso/queso/blob/dev/configure.ac#L282) and then that variable is used in the license stamping. And actually, the code that does the stamping exists in more than 1 place. How about I do a separate PR that cleans all of that up?

dmcdougall: Oh nice.  Then just leave it and we can remove it in another pull request.

m4/common/cppunit.m4
dmcdougall: Fair enough.  Thanks for clarifying.

pbauman:
Thanks!

PR #522, opened 2017-01-18
Add docs for ScopedPtr and SmartPtr

### LINKED ISSUE #397: `ScopedPtr` could fail with C++11 support
After a conversation with @roystgnr, it has occurred to us that falling back to `boost::scoped_ptr` from `std::unique_ptr` could produce code that doesn't compile.  The reason for this is that these two objects are fundamentally different.  `std::unique_ptr` permits the transfer of ownership, whereas `boost::scoped_ptr` does not.

dmcdougall:
And be clear about how shared ownership and transfer of ownership are not supported for ScopedPtr.

Fixes #397.

### DIFF
--- a/src/core/inc/ScopedPtr.h
+++ b/src/core/inc/ScopedPtr.h
@@ -35,6 +35,38 @@
 #include <memory>
 #endif
 
+/*!
+ * \file ScopedPtr.h
+ * \brief This file declares and defines a scoped pointer
+ */
+
+/*!
+ * \class ScopedPtr
+ * \brief Definition of a scoped pointer.
+ *
+ * Scoped pointers are smart pointers that clean up after themselves when they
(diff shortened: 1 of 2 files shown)


roystgnr:
You don't say anything about dying when we don't have any SharedPtr implementation.  Isn't that what will (at run time, if not compile time) happen?

dmcdougall:
> You don't say anything about dying when we don't have any SharedPtr implementation. Isn't that what will (at compile time, if not configure time) happen?

It's implied here
> If QUESO does not detect C++11 functionality, boost is required
which is a configure-time error.

dmcdougall:
I cleared it up.  Here's the added part:

> If QUESO does not detect C++11 functionality and does not detect boost then QUESO will fail to configure.

Let me know if that's better.

roystgnr:
Thanks!  :+1:

PR #526, opened 2017-01-23
Multi-dimensional Quadrature

pbauman:
The primary purpose of this PR is to deploy an interface for multi-dimensional quadrature and provide an implementation of Monte Carlo and Gaussian quadrature using tensor products of the existing one-dimensional rules. 

There was a small bit of refactoring to introduce `QuadratureBase` from which both 1D and multi-D quadrature rules subclass; this is where the quadrature weights are cached. There was some cleanup of the 1D rules (superfluous functions, using statements). I also moved `InterpolationSurrogateHelper` (and its testing) to `MultidimensionalIndexing` since we can reuse that for the `TensorProductQuadrature`.

The `TensorProductQuadrature` is unit-tested for tensor products of Legendre quadrature for a multi-dimesional linear function. The `MonteCarloQuadrature` unit test is less satisfying - the convergence is so noisy (despite the figure that's on Wikipedia...) that we can't check for a rate of 1/2, so I just check that the rate is negative. I'm open to suggestions on strengthening that test. That said, manually sampling out to 1e6+ samples, I feel confident the implementation is doing the right thing.


pbauman:
I noticed I'd accidentally turned off the 1D quadrature testing. I squashed in a fix and pushed here again.

src/misc/src/MonteCarloQuadrature.C
```diff
@@ -0,0 +1,66 @@
...
+  MonteCarloQuadrature<V,M>::MonteCarloQuadrature( const VectorSubset<V,M> & domain,
+                                                   unsigned int n_samples )
+    : MultiDQuadratureBase<V,M>(domain)
+  {
+    // We'll just uniformly sample over the domain to retrieve the quadrature
+    // points.
+    UniformVectorRV<V,M> uniform_rv("MonteCarloQuadrature_", //prefix
+                                    this->m_domain);
+
+    const BaseVectorRealizer< V, M > & 	realizer = uniform_rv.realizer();
```
dmcdougall: Did you fall asleep on your space bar?

src/misc/inc/MultiDimensionalIndexing.h
```diff
@@ -22,21 +22,21 @@
...
+#ifndef UQ_MULTI_DIMENSIONAL_INDEXING_H
+#define UQ_MULTI_DIMENSIONAL_INDEXING_H
 
 // C++
 #include <vector>
 
 namespace QUESO
 {
-  class InterpolationSurrogateHelper
+  class MultiDimensionalIndexing
```
dmcdougall: Are you concerned about this breaking older code?  If so, you might consider a simple `typedef` for a release or two before changing the name.

pbauman: Crap. I didn't think about that really. I don't think so, though, because that code was only used internally in the `InterpolationSurrogate` stuff. In the example, that class is not used at all.

src/misc/src/MonteCarloQuadrature.C
pbauman: HA! Apparently. Sorry, will fix, thanks.

src/misc/inc/1DQuadrature.h
```diff
@@ -44,8 +45,8 @@ namespace QUESO {
 
     Base class for numerical integration via quadrature rules of one-dimensional functions.
 */
-class
-Base1DQuadrature {
+class Base1DQuadrature : public QuadratureBase
```
dmcdougall: There's a name inconsistency here; `Base1DQuadrature` versus `QuadratureBase`.

dmcdougall: Is `Base1DQuadrature` intended to be abstract?  There are no pure virtual functions in here.

src/misc/src/1DQuadrature.C
```diff
@@ -33,12 +33,10 @@ Base1DQuadrature::Base1DQuadrature(
...
-  :
+  : QuadratureBase(),
   m_minDomainValue(minDomainValue),
   m_maxDomainValue(maxDomainValue),
-  m_order         (order),
-  m_positions     (0),
-  m_weights       (0)
+  m_order         (order)
 {
   queso_require_less_msg(m_minDomainValue, m_maxDomainValue, "min >= max");
```
dmcdougall: Should we check `m_maxDomainValue` is finite?

dmcdougall: Actually it appears in some cases you allow an infinite domain; `GaussianHermite1DQuadrature`.

src/misc/inc/1DQuadrature.h
dmcdougall: Actually, I notice you didn't add this class, it was already in QUESO.  So adding a pure virtual function would be a breaking change.

src/misc/src/MonteCarloQuadrature.C
```diff
@@ -0,0 +1,66 @@
...
+    this->m_positions.resize(n_samples,NULL);
+
+    for( unsigned int i = 0; i < n_samples; i++ )
+      {
+        // Create new vector from our vector space
+        // We are taking ownership of this pointer.
+        typename QUESO::SharedPtr<V>::Type domain_vec(domain.vectorSpace().newVector());
+
+        // Draw the new point
+        realizer.realization(*domain_vec);
```
dmcdougall: This will throw an exception if the volume of the domain isn't finite.  Good.

src/misc/inc/MultiDQuadratureBase.h
```diff
@@ -0,0 +1,78 @@
...
+    { queso_assert(!m_positions.empty());
+      return m_positions; }
+
+    const VectorSubset<V,M> & getDomain() const
+    { return m_domain; }
+
+  protected:
+
+    //! Locations of quadrature points
+    /*! \todo We should make this a vector of std::unique_ptr when we go full C++11. */
```
dmcdougall: I learned something today.  I always thought that you couldn't fill a `vector` with `unique_ptr`s because the contained type must satisfy the requirements of `CopyConstrubtible` and `CopyAssignable`.  That's true, but it's only true up until C++11, at which point the requirements vary depending on what operations are being applied to the container.

src/misc/inc/QuadratureBase.h
```diff
@@ -0,0 +1,60 @@
...
+
+namespace QUESO
+{
+  //! Base class for quadrature rules
+  /*! This serves as the base class for approximating integrals using numerical quadrature. Namely,
+   *   \f[ \int_{\Omega} f(x) \; dx \approx \sum_{i=1}^N f(x_i) w_i \f]
+   * This class contains the weights. Subclasses will have the positions since the position can be one-dimensional
+   * (Base1DQuadrature) or multi-dimensional (MultiDQuadratureBase).
+   */
+  class QuadratureBase
```
pbauman: Conceptually, it is abstract, but there are no virtual functions to put here. The subclass constructors will populate the weights (in this class) and the positions (in the first layer of subclasses, either 1D or multiD). Basically, this is avoiding putting weights in two different places.

We could do without 1DQuadrature entirely, and adhere to the interface in MultiDQuadratureBase, but since 1DQuadrature was already there, and its interface was specifically 1D, I went the route that I did.

src/misc/inc/1DQuadrature.h
pbauman: There used to be a pure virtual function... see c7f2820

pbauman: Do I need to scratch that OCD itch?

src/misc/inc/MultiDQuadratureBase.h
pbauman: Yeah, and it would make sense here since we never intend to transfer ownership away, so someone would have to explicitly invoke `std::move` if they tried to.

src/misc/src/MonteCarloQuadrature.C
pbauman: Yeah, I'm not sure how to handle the case when someone wants to use `MonteCarloQuadrature` for integration when one of the dimensions is infinite, e.g. when you're marginalizing a Gaussian pdf.

src/misc/inc/1DQuadrature.h
pbauman: OK, I'll do it. I guess it's not a big deal.

src/misc/inc/QuadratureBase.h
dmcdougall: If you want to prevent clients from instantiating it, make the destructor pure virtual.  I'll leave that detail up to you since it doesn't seem too important.

src/misc/inc/1DQuadrature.h
dmcdougall: I'll leave that decision to you.

dmcdougall: Hah.  Fair enough.

src/misc/src/MonteCarloQuadrature.C
dmcdougall: You don't use a uniform distribution for the realisations, you use a distribution that isn't improper on an unbounded domain and the weights are divided by PDF values at the points where you have realisations.

Thinking about it, we ought to consider other flavours of Monte Carlo quadrature with various different realiser strategies.  If you know something about the function you're integrating it might influence the choice of realiser.

dmcdougall: https://en.wikipedia.org/wiki/Importance_sampling

pbauman: *facepalm* Of course. Yes, at some point we will probably want to generalize this to do importance sampling.

dmcdougall:
Merge whenever you're happy @pbauman

src/misc/inc/QuadratureBase.h
pbauman: That's a nice idea. I didn't know about pure virtual destructors. I will do that here - I've been saved more than once by the compiler finding when I inadvertently instantiated a class that I wasn't supposed to.

pbauman:
OK, renamed `QuadratureBase` to `BaseQuadrature`, fixed the spacing issue, made quadrature base classes abstract using pure virtual destructor (TIL), and rebased on dev. Will merge once Travis is happy.

PR #527, opened 2017-01-23
Increase coverage

dmcdougall:


dmcdougall:
6.26%!  Awesome.  I think that's supposed to be percentage points, not percent.  These tests cover most of the `src/basic` directory.

src/basic/src/ScalarSequence.C
```diff
@@ -1729,14 +1729,14 @@ ScalarSequence<T>::subBasicHistogram(
   for (unsigned int j = 0; j < dataSize; ++j) {
     double value = m_seq[j];
     if (value < minHorizontalValue) {
-      bins[0] += value;
+      bins[0]++;
```
pbauman: I don't understand this change based on the commit message. Was this a bug?

pbauman: Oh wait, this change was showing up in an earlier commit too... You sure history is clean here? Or is GitHub not displaying this correctly?

pbauman: To be clear, the first commit this change shows up in is d23705cf2871e0f66810ecbfbef4092236a1e366

pbauman: The second commit this shows up in is 6f097bad449fbff7c58397edccf620130682e541

pbauman:
Modulo the comments I made, this looks good to me. I only briefly scanned the added tests, but paid closer attention to the internal QUESO changes. Great job on adding more tests. I really like CppUnit for quickly throwing in new tests.

src/basic/src/ScalarSequence.C
dmcdougall: Same bug but different functions.

`value` is the sample value.  For a histogram you don't add the value of the sample to the bin height, you add 1 to it if the sample is in that bin.  I'll update the commit message.

pbauman: OK, I see that those were different lines in the same file with copy-pasted code. Thanks for the clarification.

PR #532, opened 2017-01-25
Fixes for Gauss-Hermite Quadrature, Actually Test Gauss-Hermite Quadrature

pbauman:
In the process of adding unit tests for the Gauss-Hermite quadrature, I found two errors there (fb187e0 ac2eab6). Those errors are fixed and we have testing for Gauss-Hermite now.

### DIFF
--- a/src/misc/src/1DQuadrature.C
+++ b/src/misc/src/1DQuadrature.C
@@ -22,6 +22,8 @@
 //
 //-----------------------------------------------------------------------el-
 
+#include <sstream>
+
 #include <queso/1DQuadrature.h>
 
 namespace QUESO {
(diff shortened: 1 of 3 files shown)


dmcdougall:
```
misc/src/1DQuadrature.C:604:25: error: aggregate ‘std::stringstream ss’ has incomplete type and cannot be defined
```

pbauman:
*sigh* pushed a fix. Should I squash it into fb187e0?

pbauman:
OK, I went ahead and squashed it in.

pbauman:
OK, so I'm inferring that the codecov/patch failure is it doesn't like that only 50% of the diff ends up being tested?

That's because the other 50% is an error message. As far as I'm concerned, this is good to merge, but will wait for @dmcdougall.

dmcdougall:
Yup.

PR #528, opened 2017-01-24
Refactor Gaussian Likelihoods lnValue Interface

pbauman:
This is step 1 for refactoring/adding marginalization to the (Gaussian) likelihoods. First, renamed the base class to `LikelihoodBase` because there's nothing Gaussian about the base class. Then moved `actualValue` to the base class since all the subclasses were exactly the same function. Finally, moved to the preferred `lnValue(domainVector)` interface (from #493); along those lines, added a new `evaluateModel` interface that takes only the `domainVector` and the `modelOutput`.

PR #537, opened 2017-01-27
Fix up dregs of old lnValue API usage

dmcdougall:
Mainly in BayesianJointPdf and in ScalarFunctionSynchronizer.

I had to update some of the tests too.  It turns out that users who were
already providing gradients with the old API all of a sudden had their
implementations replaced with a finite difference.

There's a way around this behaviour that's hacky, but it's probably not
worth it unless there are serious complaints.

### DIFF
--- a/examples/gravity/src/gravity_likelihood.C
+++ b/examples/gravity/src/gravity_likelihood.C
@@ -66,8 +66,7 @@ Likelihood<V, M>::~Likelihood()
 
 template<class V, class M>
 double
-Likelihood<V, M>::lnValue(const V & domainVector, const V * domainDirection,
-    V * gradVector, M * hessianMatrix, V * hessianEffect) const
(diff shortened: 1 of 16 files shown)


dmcdougall:
This should address the test failures @pbauman was hitting while messing around with #528.

pbauman:
> There's a way around this behaviour that's hacky, but it's probably not
worth it unless there are serious complaints.

IMHO, I don't think it's worth it. There's not even really any gradient-based (or Hessian-based) examples are there? Supposedly Stochastic Newton is implemented somewhere in QUESO, but I never went fishing for it (yet). But since there are no examples, and we haven't really gotten questions about using those algorithms, I think it's very unlikely that there's any users using those algorithms and it's a lot of extra work that's not really needed.

dmcdougall:
Cool.  I'll merge then.  Let me know how #528 goes.

PR #539, opened 2017-02-02
Add Capability to Marginalize Likelihood

pbauman:
This follows up on #528 to allow the marginalization of a "marginal parameter space". That is, if instead of your model looking like f(m), it looks like f(m,q) and you want to marginalize over q, you'll now be able to do that.

The code is written, but it will be a couple of more days before I can have time to write tests. That said, i wanted to open this up now to get feedback on the interface and on the clarity of the documentation (since the integration over the marginal space can be done in a couple of ways). As such, I'm labeling as do not merge until I push some tests (and squash in any fixes for bugs identified by the testing).


dmcdougall:
That was fast.

pbauman:
:) That code was written together with #528 and #532. Instead of just letting it sit until I write the tests, I figured I could start getting feedback on the API and documentation now.

pbauman:
OK, I have two tests now. This is ready as far as I'm concerned. Feedback welcome @dmcdougall, in particular is the documentation clear? I'm a little worried the `marg_pdf_is_weight_func` in the new constructor might be confusing.

I should note that I intend to open an issue after this is merged because one thing I'd like to do is generalize this to parallelize the quadrature loop. But that's a longer discussion. As is, this is just evaluating one quadrature point at a time in the marginalization step, which is fine if your model is cheap.

pbauman:
Had forgotten to implement the new constructor for the Gaussian likelihood subclasses. Pushed in a commit that does so.

pbauman:
Labeling do not merge again. Even though my tests are passing my application that is using this capability is not behaving as expected and I suspect one or several bugs still lingers here. Will update when its sorted, one way or another.

PR #543, opened 2017-02-10
Remove emulator mean hyperprior

roystgnr:
This is more intrusive than I'd like it to be, because changing the number of hyperpriors cascades through everything, but best pull the bandaid off ASAP.


src/gp/inc/GPMSA.h
```diff
@@ -364,13 +364,16 @@ class GPMSAFactory
...
   typename ScopedPtr<VectorSpace<V, M> >::Type oneDSpace;
 
-  // Emulator mean
-  typename ScopedPtr<V>::Type emulatorMeanMin;
-  typename ScopedPtr<V>::Type emulatorMeanMax;
-  typename ScopedPtr<BoxSubset<V, M> >::Type emulatorMeanDomain;
+  // Truncation error precision
+  typename ScopedPtr<V>::Type truncationErrorPrecisionMin;
+  typename ScopedPtr<V>::Type truncationErrorPrecisionMax;
+  typename ScopedPtr<BoxSubset<V, M> >::Type truncationErrorPrecisionDomain;
```
dmcdougall: If the emulator mean is being removed altogether, then what is `truncationError*` for?

src/gp/src/GPMSA.C
```diff
@@ -1406,13 +1409,16 @@ GPMSAFactory<V, M>::setUpHyperpriors()
...
-  this->priors[1] = this->m_emulatorMean.get();
-  this->priors[2] = this->m_emulatorPrecision.get();
-  this->priors[3] = this->m_emulatorCorrelationStrength.get();
-  this->priors[4] = this->m_discrepancyPrecision.get();
-  this->priors[5] = this->m_discrepancyCorrelationStrength.get();
-  this->priors[6] = this->m_emulatorDataPrecision.get();
+  this->priors.push_back(&(this->m_parameterPrior));
+
+  if (this->num_svd_terms < numOutputs)
+    this->priors.push_back(this->m_truncationErrorPrecision.get());
```
dmcdougall: If this new `truncationErrorPrecision` term is being sampled, then shouldn't we still need the initial state for the Markov chain set, and the proposal covariance matrix element set?

test/test_Regression/test_gpmsa_cobra.C
```diff
@@ -252,22 +252,21 @@ int main(int argc, char ** argv) {
   // Initial condition of the chain
   // Have to set each of these by hand, *and* the sampler is sensitive to these
   // values
-  paramInitials[0] = 0.5; // param 1
-  paramInitials[1] = 0.5; // param 2
-  paramInitials[2] = 0.5; // param 3
-  paramInitials[3] = 0.5; // param 4
-  paramInitials[4] = 0.5; // param 5
-  paramInitials[5]  = 0.4;  // not used.  emulator mean
```
dmcdougall: This is where `truncationErrorPrecision` should go?  Or perhaps I'm missing something.

src/gp/inc/GPMSA.h
roystgnr: It's the error term for when we use fewer components of the SVD than we have outputs for.

I probably shouldn't have mixed it in this PR; I was originally intending to make more sweeping changes at once but then when I realized the regression tests would have to change I decided to "checkpoint" much earlier.

test/test_Regression/test_gpmsa_cobra.C
roystgnr: When we have enough outputs that we don't want to use every SVD term, then that's right.  I'm going to have to add some more helper/indexing APIs here, because with the "truncate based on remainder" use case, whether we have a truncation error to calibrate or not will actually be data dependent.

Issue #55, opened 2013-09-14
Calling VectorSpace copy constructor gives linker error


dmcdougall:
To reproduce:

``` c
#include <vector>
#include <string>
#include <iostream>
#include <queso/Environment.h>
#include <queso/GslVector.h>
#include <queso/GslMatrix.h>
#include <queso/VectorSpace.h>

int main(int argc, char **argv) {
  MPI_Init(&argc, &argv);

  QUESO::EnvOptionsValues options;
  options.m_numSubEnvironments = 1;

  QUESO::FullEnvironment *env = new QUESO::FullEnvironment(MPI_COMM_WORLD, "",
      "", &options);

  std::vector<std::string> names(1);
  names[0] = "my_name";
  QUESO::VectorSpace<QUESO::GslVector, QUESO::GslMatrix> vec_space(*env,
      "vec_prefix", 1, &names);
  QUESO::VectorSpace<QUESO::GslVector, QUESO::GslMatrix> vspace_cpy(vec_space);

  delete env;
  MPI_Finalize();

  return 0;
}
```

Here's the error:

``` bash
dyld: lazy symbol binding failed: Symbol not found: __ZN5QUESO9DistArrayISsE4copyERKS1_
  Referenced from: /Users/damon/git/github/queso/build/sep_impl_vectorspace/src/.libs/libqueso-0.47.dylib
  Expected in: flat namespace

dyld: Symbol not found: __ZN5QUESO9DistArrayISsE4copyERKS1_
  Referenced from: /Users/damon/git/github/queso/build/sep_impl_vectorspace/src/.libs/libqueso-0.47.dylib
  Expected in: flat namespace
```

Looks like our friend, the `DistArray` copy constructor, is back.

dmcdougall:
Linker error gone, because we replaced it with a `queso_not_implemented()`.

pbauman:
If we need to have the user not use the copy constructor, can we just make it a private member so they get a compile time error? (I haven't looked at this at all, just saw the last post in email, so I could be totally off base.)

dmcdougall:
Honestly, I'm not sure what the intent was when it was written.  The copy constructor is not private, but it calls a private function (that used to not be implemented but now is and throws a run-time error).  It's been like this for so long, and we've had no complaints about it, so maybe we ought to just make the copy constructor private and remove the implementation altogether.  I'll open a PR for it.

PR #547, opened 2017-02-23
Make copy ctor for DistArray and VectorSpace private

### LINKED ISSUE #57: Test VectorSpace copy constructor
This relies on the `copy` method in `DistArray` being implemented.

dmcdougall:
They were throwing exceptions at run-time anyway, so clearly nobody was using them.

This addresses a comment made by @pbauman in #55, and fixes #57.


pbauman:
Fine with me. Can revisit if someone decides they need to copy the vector space.

pbauman:
What is the purpose of `DistArray` BTW? Where is that used?

dmcdougall:
> What is the purpose of DistArray BTW? Where is that used?

I have no idea.

dmcdougall:
The failure was a decrease in coverage, which I expected.

PR #548, opened 2017-02-23
Fix trilinos communicator in serial

### LINKED ISSUE #458: test_serialEnv fails when Trilinos is enabled
We get e.g.: "Attempting to use an MPI routine before initializing MPICH", from the Epetra_MpiComm constructor via the QUESO::MpiComm constructor.

dmcdougall:
Fixes #458.

@roystgnr Can you take a look at this?  Not happy with the dynamic cast.

### DIFF
--- a/src/core/inc/MpiComm.h
+++ b/src/core/inc/MpiComm.h
@@ -27,12 +27,12 @@
 
 #include <queso/Defines.h>
 
-#ifdef QUESO_HAS_TRILINOS
-#include <Epetra_MpiComm.h>
-#endif
-
 #ifdef QUESO_HAS_MPI
 #include <mpi.h>
+#endif
+
+#ifdef QUESO_HAS_TRILINOS
+class Epetra_Comm;
 #endif
 
 namespace QUESO {
--- a/src/core/inc/MpiComm.h
+++ b/src/core/inc/MpiComm.h
@@ -242,8 +242,10 @@ public:
 
   //! @name Attribute Accessor Methods
   //@{
+#ifdef QUESO_HAS_MPI
   //! Extract MPI Communicator from a MpiComm object.
   RawType_MPI_Comm Comm     () const;
+#endif  // QUESO_HAS_MPI
 
   //! Return my process ID.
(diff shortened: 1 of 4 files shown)


roystgnr:
Options other than dynamic_cast:

Something like libMesh::cast_ptr, which does dynamic_cast when debugging but a faster static_cast in opt or with RTTI disabled.  In this case the performance difference would be undetectable and this would not be an improvement.

Keep both base and derived class pointers to the same object.  You'd be basically reinventing the RTTI wheel with that.

You suggested putting the whole Comm() method inside the QUESO_HAS_MPI ifdefs?  That doesn't get rid of the dynamic_cast, but it is a good idea IMHO.  If the user ever has to ask for a raw MPIComm even *with* MPI enabled then there's not enough serial-vs-parallel encapsulation going on, and it'd be nice to find that out at compile time rather than run time.

roystgnr:
Whoa, don't misparaphrase me in your commit logs!

Having the raw MPIComm available is good.  We definitely want that method to exist, in case there's functionality we failed to expose or third-party APIs which require that specific type.  It's just if we find ourselves ever *using* that method internally, that would be a sign of a flaw.

dmcdougall:
Apologies.  I'll update the commit message.

dmcdougall:
>Something like libMesh::cast_ptr, which does dynamic_cast when debugging but a faster static_cast in opt or with RTTI disabled. In this case the performance difference would be undetectable and this would not be an improvement.

Agreed.  For now, I'll keep the dynamic cast since I don't expect this method to be used in a tight loop.  Can make the transition to `queso_cast` once I find enough time to borrow *cough* steal *cough* libmesh's macros.

>You suggested putting the whole Comm() method inside the QUESO_HAS_MPI ifdefs? That doesn't get rid of the dynamic_cast, but it is a good idea IMHO. If the user ever has to ask for a raw MPIComm even with MPI enabled then there's not enough serial-vs-parallel encapsulation going on, and it'd be nice to find that out at compile time rather than run time.

Done.

PR #550, opened 2017-03-30
New solveWithBayesMetropolisHastings methods

dmcdougall:
I added some new `solve` methods.

Before this change, the user was required to pass the options object, initial point, and proposal covariance matrix.

After this change, there are new methods that will allow you to omit this information.  I haven't implemented every permutation of these, but these should be enough for now:

```c
// Uses prior variance for proposal cov matrix
void solveWithBayesMetropolisHastings(const MhOptionsValues * alternativeOptionsValues, const P_V & initialValues);
// Uses prior variance and prior mean for proposal cov matrix and initial point
void solveWithBayesMetropolisHastings(const MhOptionsValues * alternativeOptionsValues);
// Uses default options, prior mean, and prior covariance for all parameters
void solveWithBayesMetropolisHastings();
```

__Edit__: I accidentally a word.


dmcdougall:
I believe this was one of the CASL asks from ages ago.  Work from @roystgnr made this a lot easier to do.  Thanks @roystgnr!

Issue #555, opened 2017-04-13
Re-evaluate what space adaptivity happens in with the logit transform turned on


dmcdougall:
It appears that adaptive metropolis happens (computing sample covariance matrix) in the [transformed (unbounded) space](https://github.com/libqueso/queso/blob/14aefed7b6a56a9de8f8b2b5b1a7a8d23955f300/src/stats/src/MetropolisHastingsSG.C#L2043-L2058).  That is, we compute the sample covariance matrix of logit-transformed samples and update the TK with this matrix.

This approach means we have some logit-transform logic that lies *outside* of the transition kernel and in the sequence generator.  If, instead, we do the adaptive metropolis procedure in user-space (on samples that have not been logit-transformed), then one could conceivably hide all the transform logic in the transition kernel instead.

Issue #472, opened 2016-10-28
Metropolis Hastings fails when param domain is a ConcatenationSubset


moprak:
The failure happens inside `MetropolisHastingsSG<P_V, P_M>::commonConstructor()`. If I'm not mistaken, this happens at line 544 in v0.55.0 brach in MetropolisHastingsSG.C

`dynamic_cast<const BoxSubset<P_V, P_M> & >(m_targetPdf.domainSet()), drScalesAll, m_initialProposalCovMatrix);`

My param domain is a `QUESO::ConcatenationSubset`. I don't know enough to understand why this should or shouldn't fail though.

moprak:
I can avoid hitting this if I set `ip_mh_doLogitTransform = 0` in the input file.

PR #556, opened 2017-04-13
Refactor logit transform logic into TK

dmcdougall:
The fact we were doing logit transforms outside of the TK designed for logit transforms bothered me.

I can't do it everywhere, though, one of the pieces of logic relies fundamentally on the adaptivity formulation.  See #555 for details.

### DIFF
--- a/src/stats/inc/TKGroup.h
+++ b/src/stats/inc/TKGroup.h
@@ -115,6 +115,14 @@ public:
    * After this method is called, covMatrixIsDirty should return \c true.
    */
   virtual void cleanCovMatrix() = 0;
+
+  //! Sets and proposal covariance matrix (and for all DR stage RVs as well)
+  /*!
+   * If the TK has a proposal covariance matrix, set the covariance matrix.
+   * If the TK support delayed rejection, derived classes should scale by a
+   * factor of \f$ 1/scales^2 \f$.
+   */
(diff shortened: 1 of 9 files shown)


src/stats/inc/TKGroup.h
```diff
@@ -115,6 +115,14 @@ class BaseTKGroup {
...
    */
   virtual void cleanCovMatrix() = 0;
+
+  //! Sets and proposal covariance matrix (and for all DR stage RVs as well)
+  /*!
+   * If the TK has a proposal covariance matrix, set the covariance matrix.
+   * If the TK support delayed rejection, derived classes should scale by a
+   * factor of \f$ 1/scales^2 \f$.
+   */
+  virtual void updateLawCovMatrix(const M & covMatrix) = 0;
```
dmcdougall: Note, this change break backwards compatibility.  In my defence, I highly doubt anybody is subclassing TKs yet; the factory infrastructure is quite new.

Instead, I can make this function virtual but non-pure and throw a run-time exception instead.

dmcdougall:
This PR serves as a first step in addressing #472.

src/stats/inc/TKGroup.h
roystgnr: I'd definitely prefer the compile-time failure option to the run-time failure option...

roystgnr:
:+1:

PR #557, opened 2017-04-14
Make logit work on more general (non-BoxSubset) domains

### LINKED ISSUE #472: Metropolis Hastings fails when param domain is a ConcatenationSubset

dmcdougall:
Fixes #472.

### DIFF
--- a/src/basic/inc/BoxSubset.h
+++ b/src/basic/inc/BoxSubset.h
@@ -25,7 +25,6 @@
 #ifndef UQ_BOX_SUBSET_H
 #define UQ_BOX_SUBSET_H
 
-#include <queso/VectorSpace.h>
 #include <queso/VectorSubset.h>
 
 namespace QUESO {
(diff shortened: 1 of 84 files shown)


dmcdougall:
#556 should be merged before this.

dmcdougall:
@roystgnr Take a look?

roystgnr:
:+1: to merge as-is.

There's probably a bunch of other code we can fix afterward, too:
```
$ srcgrep dynamic_cast | grep BoxSubset | wc -l
14
```

A few cases seem to be truly assuming a box domain in the code which follows, but other cases could handle general domains if only we switched a mins/maxes query to the new API.

PR #560, opened 2017-04-19
GCC-4.4.7 Compatibility: explicit SharedPtr() ctor

briadam:
Change a couple instances of NULL to SharedPtr<>::Type() to satisfy
older compiler.  There are probably many ways to fix this; feel free to choose another.


dmcdougall:
Thank you!

PR #564, opened 2017-04-20
Pimpl hdf5

### LINKED ISSUE #256: Move out hdf5 header from Environment.h
In similar spirit to #240, I have an app using QUESO that is not, locally, using HDF5. But, because of this header, I'm going to need to include it in the build of my app. Which is weak.

dmcdougall:
This fixes #256.

It's technically a breaking change; I'm moving the `FilePtrSetStruct` out to a header file that isn't installed on the user's system.  I don't expect that user's are using QUESO's file-handling capabilities.


dmcdougall:
@pbauman Let me know what you think.

PR #566, opened 2017-04-21
Gpmsa input parameter scaling

roystgnr:
This PR adds input file options and APIs for normalization of both uncertain parameters and scenario parameters in the GPMSA code.


roystgnr:
@briadam - this is the PR I was referring to in this afternoon's email.

dmcdougall:
`ERROR: ./test_gpmsa/test_gpmsa_autoscaled_samples.m file does not exist`

src/gp/src/GPMSAOptions.C
```diff
@@ -271,6 +297,293 @@ GPMSAOptions::~GPMSAOptions()
...
+          varTerm *= varTerm;
+          varScenario += varTerm;
+        }
+
+      for (unsigned int i=0; i < m_experimentScenarios.size(); ++i)
+        {
+          V varTerm = *m_experimentScenarios[i] - meanScenario;
+          varTerm *= varTerm;
+          varScenario += varTerm;
+        }
```
dmcdougall: These routines for calculating the variance aren't numerically stable.  You can use [this](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Online_algorithm) instead.

src/gp/src/GPMSAOptions.C
```diff
@@ -271,6 +297,293 @@ GPMSAOptions::~GPMSAOptions()
...
+          varTerm *= varTerm;
+          varUncertain += varTerm;
+        }
+
+      for (unsigned int i=0; i < m_experimentScenarios.size(); ++i)
+        {
+          V varTerm = *m_experimentScenarios[i] - meanUncertain;
+          varTerm *= varTerm;
+          varUncertain += varTerm;
+        }
```
dmcdougall: Ditto.  See above for a numerically stable procedure.

src/gp/inc/GPMSAOptions.h
```diff
@@ -86,6 +87,104 @@ class GPMSAOptions
...
+  //! Do automatic normalization, using mean and variance of the
+  //  supplied data, for all input uncertain parameters, all input
+  //  scenario parameters, and all output values.
+  //
+  //  Normalized values will have mean 0 and standard deviation 1.
+  void set_autoscale_meanvar();
+
+  //! Set a value, for uncertain parameter i in simulation inputs, of
+  //  the physical parameter range (range_min, range_max) which should
+  //  be rescaled to (0,1)
```
dmcdougall: Ah, ok.  So we can use our own range or use the min/max of the supplied data.  It took me a minute to realise the user may want to specify a range that isn't present in the data.  Might we check that, for example, `range_max` is larger than the max of the data?

src/gp/src/GPMSAOptions.C
```diff
@@ -271,6 +297,293 @@ GPMSAOptions::~GPMSAOptions()
...
+                                      (*m_experimentScenarios[i])[p]);
+            maxScenario[p] = std::max(maxScenario[p],
+                                      (*m_experimentScenarios[i])[p]);
+          }
+
+      for (unsigned int p=0; p != dimScenario; ++p)
+        if (m_autoscaleMinMaxAll ||
+            m_autoscaleMinMaxScenario.count(p))
+          {
+            if ((m_scenarioScaleMin.size() > p) &&
```
dmcdougall: `m_scenarioScaleMin`, if nonempty, is always of size `dimScenario`, right?  So this part of the `if` statement is only `false` if it's empty?

src/gp/src/GPMSAOptions.C
```diff
@@ -271,6 +297,293 @@ GPMSAOptions::~GPMSAOptions()
...
+                ((m_scenarioScaleMin[p] != 0) ||
+                 (m_scenarioScaleRange[p] != 1)))
+              queso_error_msg("Cannot autoscale and manually scale the same scenario parameter");
+
+            this->set_scenario_parameter_scaling(p, minScenario[p],
+                                                    maxScenario[p]);
```
dmcdougall: Will the user ever call this?

src/gp/inc/GPMSAOptions.h
dmcdougall: It took me a minute to realise why we need to allow a general range.  It's because the user may want to specify a range besides the min/max of the supplied data.

src/gp/inc/GPMSAOptions.h
```diff
@@ -86,6 +87,104 @@ class GPMSAOptions
...
+  //
+  //  If no value is set and no automatic normalization is specified,
+  //  a default of (0,1) will be used; i.e. no normalization.
+  //
+  //  This option is mutually exclusive with the above automatic
+  //  scaling options; trying to use both is a runtime error.
+  void set_uncertain_parameter_scaling(unsigned int i,
```
dmcdougall: So, the user can call this and set their own range, right?

src/gp/src/GPMSAOptions.C
```diff
@@ -271,6 +297,293 @@ GPMSAOptions::~GPMSAOptions()
...
+   const std::vector<typename SharedPtr<V>::Type> & m_experimentScenarios,
+   const std::vector<typename SharedPtr<V>::Type> & m_experimentOutputs)
+{
+  if ((m_autoscaleMinMaxAll && m_autoscaleMeanVarAll) ||
+      ((m_autoscaleMinMaxAll || m_autoscaleMeanVarAll) &&
+       (!m_autoscaleMinMaxUncertain.empty() ||
+        !m_autoscaleMeanVarUncertain.empty() ||
+        !m_autoscaleMinMaxScenario.empty() ||
+        !m_autoscaleMeanVarScenario.empty())))
+    queso_error_msg("Cannot autoscale based on incompatible criteria");
```
dmcdougall: Ok, cool, I understand this.

src/gp/src/GPMSAOptions.C
```diff
@@ -271,6 +297,293 @@ GPMSAOptions::~GPMSAOptions()
...
+      for (unsigned int p=0; p != dimScenario; ++p)
+        if (m_autoscaleMinMaxAll ||
+            m_autoscaleMinMaxScenario.count(p))
+          {
+            if ((m_scenarioScaleMin.size() > p) &&
+                ((m_scenarioScaleMin[p] != 0) ||
+                 (m_scenarioScaleRange[p] != 1)))
+              queso_error_msg("Cannot autoscale and manually scale the same scenario parameter");
+
+            this->set_scenario_parameter_scaling(p, minScenario[p],
```
dmcdougall: Hmm, is it also the case the user can call `set_scenario_parameter_scaling` and set their own range, or is this method really only supposed to be called by the `GPMSAOptions` object?

src/gp/src/GPMSAOptions.C
dmcdougall: Unfortunately, this procedure for calculating the variance isn't numerically stable.  You can use [this](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Online_algorithm) instead, which is stable.

src/gp/src/GPMSAOptions.C
dmcdougall: See my other comment about stability.

dmcdougall:
So I learned an important lesson when writing this feedback.  If you're reviewing a PR and you leave a bunch of comments, and then part-way through you decide you'd like know if a method the PR added is `public:` or `private:` and you need to click the `view` button to see the whole file, be warned that all your well thought out comments will disappear into the ether.  Fucking unbelievable.  I wasted an hour of my time.

@roystgnr Let me know if you have any questions.

dmcdougall:
Actually, I retract my previous statement.  Comments don't disappear into the ether.  They are submitted wholesale without your knowledge.

dmcdougall:
I apologise for my French.

src/gp/src/GPMSAOptions.C
roystgnr: After scaling is finalized, it will be of size dimScenario.

Beforehand, it's possible for users to set a scaling manually for a particular index, in which case we'll have resized to hold that index but not to hold every index.

src/gp/src/GPMSAOptions.C
roystgnr: This is designed for users to use too.

src/gp/inc/GPMSAOptions.h
roystgnr: Nah.  If the user wants to scale using mean + variance, for example, it's guaranteed that they'll have data outside [0, +1SD].

As far as GPMSA is concerned, your data really ought to be nondimensionalized, but there's probably room to experiment with what *kind* of nondimensionalization works best.

src/gp/inc/GPMSAOptions.h
roystgnr: Right.

src/gp/src/GPMSAOptions.C
```diff
@@ -271,6 +336,246 @@ GPMSAOptions::~GPMSAOptions()
...
+
+      for (unsigned int i=0; i < m_simulationScenarios.size(); ++i)
+        mean_var_update(n, meanScenario, varScenario,
+                        *m_simulationScenarios[i]);
+
+      for (unsigned int i=0; i < m_experimentScenarios.size(); ++i)
+        mean_var_update(n, meanScenario, varScenario,
+                        *m_experimentScenarios[i]);
+
+      varScenario /= n;
```
dmcdougall: Ah, because `mean_var_update` changes `n`.  Ok.

briadam:
A couple usability questions on this.  If a client requests domain and/or data scaling and then queries the inverse problem and sequence generator for their data after calibration, will that be in the scaled or original space?  Presuming scaled, are there utilities to transform the resulting posterior chain back (and any other relevant solution data) to unscaled space?  If not, I might have to retain scaling on the Dakota side, so I can invert it when reporting results to users...  What will a client get back if asking for the parameter domain or data after specifying the scaling options?

dmcdougall:
Thanks @briadam.  I believe the results presented to the user will not be scaled back to the original space.  @roystgnr will surely correct me if I'm wrong.

dmcdougall:
(And this pull request doesn't provide inverse mappings.)

roystgnr:
Output data is in the original, physical space.  We do the scaling entirely internal to the GPMSA class, so your priors are still specified in physical space and your calibrated data is still returned in physical space.

This does mean that the *benefit* of scaling is also entirely internal to the GPMSA class: you no longer have dimensional insanity in your covariance terms.  If scaling would also be beneficial to the inverse problem solver then that'll be a separate issue... but I think the same "deal with it internally" philosophy ought to apply.  Returning rescaled data to the user just seems cruel and error-prone.

briadam:
Okay, I see, that sounds great to me.  I like it!

dmcdougall:
I'm not sure I understand.  If the user doesn't normalise their data, and opts to have QUESO do it, I don't think the calibrated parameters (and hyperparameters) are mapped back to the user's physical space.

Look here: https://github.com/libqueso/queso/pull/566/files#diff-8a60952749a0bb180833223a9b205628R274.  Notice that the scenario and uncertain parameters are normalised, but `emulator_corr_strength` isn't touched at all.

Am I missing something?

roystgnr:
Notice that that's not a multiplication or an addition, it's an exponentiation.  The exponent is by definition unitless.  So in a sense, "If the user doesn't normalize their data" is impossible - if they don't normalize, they are implicitly normalizing by their choice of unit.

Also, notice, those normalization calls are *internal*.  GPMSA expects the inverse problem solver to hand it a physical parameter, and the inverse problem solver doesn't ever see the normalization.

dmcdougall:
> GPMSA expects the inverse problem solver to hand it a physical parameter, and the inverse problem solver doesn't ever see the normalization.

Ok, I understand this.  The normalisation is a change of variables, though, so I'd still expect to see a Jacobian calculation.

dmcdougall:
Nevermind, it's a linear transformation so the Jacobian is a constant.  I'm satiated.

briadam:
One more question on this as I integrate into Dakota. Not based on runtime, but a quick read through the code, it appears that if there is no variability in the scenario parameters, e.g., when we're mocking up a single scenario where they are all 0.5 for the case where there are no scenario variables, a divisor of 0 will be calculated (min = max, so range or variance = 0).  Is my concern founded, or are there guards somewhere for degenerate cases?

For this case, perhaps instead of using the autoscale feature, I'll have to use the finer-grained API to scale the uncertain parameters, but not the scenario parameters... In fact, even for a user-provided scenario case, I'd have to check that they are not all identical to avoid an issue here...

roystgnr:
Yes, I'm afraid that's the case: right now if you try to autoscale and the data is all at one value, you'll get NaNs.  I'll fix that shortly but the fix won't be backported.

Note that there's actually two levels of finer-grained API, though: in addition to being able to say "here's the min and max to scale parameter p to", you can also say "autoscale parameter p", even if you're not going to autoscale parameter p+1.

briadam:
Okay, thanks for clarifying.  Yes, saw those two APIs.  In starting to use them to differentiate calibration vs. scenario scaling, I was originally worried that there wasn't a fine-grained API to toggle data scaling separately from parameter scaling, then I noticed data scaling isn't yet implemented, so that's not a limitation... I'll keep the data scaling implementation on my client side for now.

PR #571, opened 2017-04-24
Fix incorrect variable name in GPMSA GetPot parse

dmcdougall:


roystgnr:
:+1:

After looking at the diff, I'm pretty sure the "review request" from @dmcdougall was more of a "public shaming invitation".

dmcdougall:
Honestly, it wasn't intended to shame at all.

dmcdougall:
Will merge once travis is happy.

roystgnr:
Just kidding.  There's no way to make that level of copy-and-paste error less shameful.

roystgnr:
If we weren't planning on removing the boost parsing option eventually anyway, I'd say we need to handle these cases at runtime... let someone build with both boost and GetPot, make the choice of which to use selectable via e.g. command line option, and run our regression tests with both every time.

PR #572, opened 2017-04-24
GCC 4.4.7 compatibility: remove typename in GPMSAOptions.C

briadam:
(Dakota won't support this compiler much longer...)


dmcdougall:
@briadam Can you target the `hotfix-0.57.1` branch instead?

briadam:
Yes, I wondered if that would be easier for your workflow.  I don't see a way to edit a pull request that way, so I'll close and make new one...

PR #573, opened 2017-04-24
GCC 4.4.7 compatibility: remove typename in GPMSAOptions.C

briadam:
(Dakota won't support this compiler much longer...)


briadam:
I probably did more work than was needed to apply my changes to hotfix-0.57.1 when I could have just re-targeted the pull request?  If I messed up anything with the history, let me know and I can try again.

PR #574, opened 2017-04-24
Make boost optional, allows `--without-boost` config

dmcdougall:
Not sure on the interaction between this and lack of C++11.

@pbauman I believe this is relevant to you:

It's happening


dmcdougall:
We have boost installed in `/usr/local` on the work machines, which made this a pain to test.  I think I've got it all figured out now.

Issue #575, opened 2017-04-24
GPMSA with shifted non [0,1] prior


briadam:
Per discussion with @roystgnr and @dmcdougall I believe the attached variant of test_gpmsa_vector.C reproduces the issue I'm seeing from Dakota.  The calibration domain appears wrong in the aggregate GPMSA prior, and something downstream breaks.  Yet the distributionMean gives the right initial point...

I expect the calibration to be performed over the [-1, 0] hypercube for the calibration parameters.  What I see instead is (truncated):

```
BMA: priorRv PDF domainSet on [-1, 0]
In BoxSubset<V,M>::print(): m_minValues = -1 -1 -1 -1 -1 , m_maxValues = 0 0 0 0 0 , m_volume = 1

BMA: GPMSA prior PDF domainSet (calib params now on [0,1]!?!
In BoxSubset<V,M>::print(): m_minValues = 0 0 0 0 0 0.3 0.3 0.3 0 0 0 0 0 0 0 0 60 0.3 , m_maxValues = 1 1 1 1 1 inf inf inf 1 1 1 1 1 1 inf 1 100000 inf , m_volume = inf
```

Yet somehow the distributionMean is still on [-1.0, 0.0].  And despite the initial values appearing to be between those bounds...

```
BMA: Default initials
-0.5 -0.5 -0.5 -0.5 -0.5 1 1 1 0.909091 0.909091 0.909091 0.909091 0.909091 0.909091 10000 0.909091 999.999 1 
BMA: Final initials
-0.5 -0.5 -0.5 -0.5 -0.5 0.4 0.4 0.4 0.97 0.97 0.97 0.97 0.2 0.8 10 0.97 8000 1

Assertion `!(outOfTargetSupport)' failed.
initial position should not be out of target pdf support
../../src/stats/src/MetropolisHastingsSG.C, line 1405, compiled Apr 19 2017 at 09:29:14
```

[test_gpmsa_vector.C.txt](https://github.com/libqueso/queso/files/953107/test_gpmsa_vector.C.txt)

briadam:
Note that my example doesn't shift the simulation data to [-1, 0], so the example isn't complete.  I was just trying to demonstrate what might be an issue with the prior...

[...]

PR #577, opened 2017-04-27
GPMSA should respect existing parameter domain

roystgnr:
This is backwards compatible with the existing unit tests for me, but also fixes the broken test case that @briadam proposed in #575

### DIFF
--- a/src/gp/inc/GPMSA.h
+++ b/src/gp/inc/GPMSA.h
@@ -32,6 +32,7 @@
 #include <queso/VectorSpace.h>
 #include <queso/VectorRV.h>
 #include <queso/ConcatenatedVectorRV.h>
+#include <queso/ConcatenationSubset.h>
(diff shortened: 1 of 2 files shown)


roystgnr:
@briadam, I know it's too late and too big a dependency list to get this in a bugfix release, but if you could try it out I'd appreciate it.  Damon is deep in SciSoftDays right now and won't be glancing at QUESO until Monday.

dmcdougall:
This looks fine to me.  I can't believe I didn't think of `.imageSet()` when I initially wrote this.

Should I merge or wait for input from @briadam?

briadam:
Thanks Damon. I reviewed briefly last week, but might not be able to test until after today's code freeze. If I recall, this is targeting dev (can't load right now), so the timing with our release isn't critical. Okay to merge by me.

Brian
________________________________

PR #584, opened 2017-05-25
Fix compiler warnings 

dmcdougall:
QUESO now compiles and tests cleanly with

```
CXXFLAGS="-O0 -felide-constructors -g -pedantic -W -Wall -Wextra -Wno-long-long -Wunused -Wpointer-arith -Wformat -Wparentheses -Woverloaded-virtual"
```

This was pretty painful to do.

### DIFF
--- a/examples/gp/pseudovector/gpmsa_pseudovector.C
+++ b/examples/gp/pseudovector/gpmsa_pseudovector.C
@@ -39,7 +39,7 @@ readData(const std::vector<QUESO::SharedPtr<QUESO::GslVector>::Type> & simulatio
     queso_error_msg("dakota_pstudy.dat was unreadable");
 
   i = 0;
-  while (fscanf(fp_in, "%d %lf %lf %lf %lf %lf %lf\n", &id, &k_tmasl, &k_tmoml,
+  while (fscanf(fp_in, "%u %lf %lf %lf %lf %lf %lf\n", &id, &k_tmasl, &k_tmoml,
         &k_tnrgl, &k_xkwlx, &k_cd, &pressure) != EOF) {
     (*(simulationScenarios[i]))[0] = 0.5;
 
--- a/examples/gp/pseudovector/gpmsa_pseudovector.C
+++ b/examples/gp/pseudovector/gpmsa_pseudovector.C
@@ -12,7 +12,7 @@ std::vector<double>
 readData(const std::vector<QUESO::SharedPtr<QUESO::GslVector>::Type> & simulationScenarios,
          const std::vector<QUESO::SharedPtr<QUESO::GslVector>::Type> & simulationParameters,
          const std::vector<QUESO::SharedPtr<QUESO::GslVector>::Type> & simulationOutputs,
-         const std::vector<QUESO::SharedPtr<QUESO::GslVector>::Type> & experimentScenarios,
+         const std::vector<QUESO::SharedPtr<QUESO::GslVector>::Type> & /* experimentScenarios */,
          const std::vector<QUESO::SharedPtr<QUESO::GslVector>::Type> & experimentOutputs) {
   FILE * fp_in = fopen("gp/vector/dakota_pstudy.dat", "r");
   if (!fp_in)
--- a/examples/gp/pseudovector/gpmsa_pseudovector.C
+++ b/examples/gp/pseudovector/gpmsa_pseudovector.C
@@ -7,6 +7,8 @@
 
 #include <cstdio>
 
+#define LINE_SIZE 512
+
 // Read in data files.  Return std deviations for each output.
 std::vector<double>
 readData(const std::vector<QUESO::SharedPtr<QUESO::GslVector>::Type> & simulationScenarios,
@@ -20,9 +22,9 @@ readData(const std::vector<QUESO::SharedPtr<QUESO::GslVector>::Type> & simulatio
   if (!fp_in)
     queso_error_msg("Cannot find dakota_pstudy.dat");
 
-  unsigned int i, id, size = 512;
+  unsigned int i, id;
   double k_tmasl, k_tmoml, k_tnrgl, k_xkwlx, k_cd, pressure;
-  char line[size];
+  char line[LINE_SIZE];
 
   // The user should know what the bounds on the data are
   double mins[] = {0.95, 0.9, 0.9, 0.9, 0.9};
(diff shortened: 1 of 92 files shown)


roystgnr:
In principle I strongly approve of this PR.  I can't promise to review all 86 commits, though...

dmcdougall:
> In principle I strongly approve of this PR. I can't promise to review all 86 commits, though...

Yeah, sorry.  I did it that way because that's how I did it locally; fix up one file at a time.  I can squash all the "comment out unused parameters" commits together.

examples/gp/pseudovector/gpmsa_pseudovector.C
```diff
@@ -7,22 +7,24 @@
...
   FILE * fp_in = fopen("gp/vector/dakota_pstudy.dat", "r");
   if (!fp_in)
     fp_in = fopen("dakota_pstudy.dat", "r");
   if (!fp_in)
     queso_error_msg("Cannot find dakota_pstudy.dat");
 
-  unsigned int i, id, size = 512;
+  unsigned int i, id;
   double k_tmasl, k_tmoml, k_tnrgl, k_xkwlx, k_cd, pressure;
-  char line[size];
```
roystgnr: Looks this didn't make it from C99 into C++ until C++14, but we've been skating by on compiler-specific extensions?

src/contrib/getpot/getpot.h
```diff
@@ -532,7 +532,7 @@ class GetPot
    * multiple threads at once, so we'll wrap access to
    * mutable objects in a mutex.
    */
-  GETPOT_MUTEX_DECLARE;
+  GETPOT_MUTEX_DECLARE
```
roystgnr: Wait, why are we doing this?

src/core/src/GslMatrix.C
```diff
@@ -1881,7 +1881,8 @@ GslMatrix::subReadContents(
     unsigned int posInTmpString = 6;
 
     // Isolate 'n_rows' in a string
-    char nRowsString[tmpString.size()-posInTmpString+1];
+    std::string nRowsString;
+    nRowsString.reserve(tmpString.size()-posInTmpString+1);
```
roystgnr: Pretty sure the C++ standard says you're playing with undefined behavior here:
http://www.cplusplus.com/reference/string/basic_string/operator[]/

> If pos is less or equal to the string length, the function never throws exceptions (no-throw guarantee).
> Otherwise, it causes undefined behavior.

Increasing the string capacity allows internal optimizations, but doesn't increase the string length.

src/core/src/GslMatrix.C
```diff
@@ -1891,7 +1892,8 @@ GslMatrix::subReadContents(
 
     // Isolate 'n_cols' in a string
     posInTmpString++; // Avoid reading ',' char
-    char nColsString[tmpString.size()-posInTmpString+1];
+    std::string nColsString;
+    nColsString.reserve(tmpString.size()-posInTmpString+1);
```
roystgnr: Likewise here.

src/misc/src/Miscellaneous.C
```diff
@@ -53,12 +53,13 @@ MiscReadDoublesFromString(
     if (inputString[i] == ' ') {
       if (aDoubleIsBeingRead == true) {
         // We just finished reading the current string/double. Convert string to double now.
-        char tmpVar[numberOfChars+1];
+        std::string tmpVar;
+        tmpVar.reserve(numberOfChars+1);
         for (std::string::size_type j = 0; j < numberOfChars; ++j) {
```
roystgnr: Likewise here.

src/misc/src/Miscellaneous.C
```diff
@@ -76,12 +77,13 @@ MiscReadDoublesFromString(
   } // for
   if (aDoubleIsBeingRead == true) {
     // We just finished reading the current string/double. Convert string to double now.
-    char tmpVar[numberOfChars+1];
+    std::string tmpVar;
+    tmpVar.reserve(numberOfChars+1);
     for (std::string::size_type j = 0; j < numberOfChars; ++j) {
```
roystgnr: And here.

src/misc/src/Miscellaneous.C
```diff
@@ -105,12 +107,13 @@ MiscReadWordsFromString(
     if (inputString[i] == ' ') {
       if (aWordIsBeingRead == true) {
         // We just finished reading the current string/word.
-        char tmpVar[numberOfChars+1];
+        std::string tmpVar;
+        tmpVar.reserve(numberOfChars+1);
```
roystgnr: I'm now noticing a pattern.

src/misc/src/Miscellaneous.C
```diff
@@ -128,12 +131,13 @@ MiscReadWordsFromString(
   } // for
   if (aWordIsBeingRead == true) {
     // We just finished reading the current string/word.
-    char tmpVar[numberOfChars+1];
+    std::string tmpVar;
+    tmpVar.reserve(numberOfChars+1);
```
roystgnr: Wonder if I'll end up with more comments than altered files?

src/contrib/getpot/getpot.h
dmcdougall: When I was compiling I didn't have libmesh, so that macro was empty.  `-Wpedantic` complains when you have a single stray semicolon, so I put the semicolon inside the macro and left it off here.

examples/gp/pseudovector/gpmsa_pseudovector.C
dmcdougall: Yes.

src/core/src/GslMatrix.C
dmcdougall: Ok, Then I'll initialise it with `tmpString.size()-posInTmpString+1` spaces.

src/contrib/getpot/getpot.h
roystgnr: Okay, I get it now.  Stray semicolons are fine for -Wpedantic (and would be easy to work around if they weren't) in function/method bodies, but not in struct layouts.

And I can't think of any workaround that avoids the warning and keeps the same syntax but doesn't pointlessly pad the struct in the !USE_LIBMESH case, so I guess this is fine.

roystgnr:
:+1: after you fix the strings.  Initializing with spaces sounds fine.

PR #590, opened 2017-06-07
Make QUESO define GETPOT_NAMESPACE only if needed

dmcdougall:
And `#undef` when we're done with the #include.

@pbauman You should do this in `GRINS` too.  You don't do this in `GRINS` header files, and hence the getpot clash.  From the grins top-level directory:

```
$ git grep -l "#include \"libmesh/getpot\.h\"" | wc -l
139
```

Libmesh doesn't ever include getpot in their headers, so they don't need it.  I just did it everywhere, including in implementation files, just to be explicit.

### DIFF
--- a/src/basic/src/ScalarFunction.C
+++ b/src/basic/src/ScalarFunction.C
@@ -32,7 +32,9 @@
 #ifndef QUESO_DISABLE_BOOST_PROGRAM_OPTIONS
 #include <queso/BoostInputOptionsParser.h>
 #else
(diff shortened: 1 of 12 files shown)


roystgnr:
This works as long as everyone does it.  Is there any way to save any pre-existing macro value then restore it, so we'd be compatible with other code that *isn't* politely undefining GETPOT_NAMESPACE?

dmcdougall:
[Doesn't appear so](https://stackoverflow.com/questions/14261534/temporarily-overwrite-a-macro-in-c-preprocessor).

PR #593, opened 2017-06-11
Update travis

dmcdougall:
Make travis do an out-of-source build, a distcheck, run on newer ubuntu distribution, and dump test suite output to screen.

This'll hopefully catch tarball problems before they're released into the wild.


roystgnr:
Looks good

Issue #511, opened 2017-01-10
Fix lambda_eta vs lambda_wi distinction


roystgnr:
GPMSA users need to be able to specify independent hyperpriors for the two.

When we're in a multivariate or functional case with a non-truncated SVD, truncation_error_precision (lambda_eta in the paper's multivariate case) needs to go unused!

In the near future we can leave truncation_error_parameter as a "dummy" hyperparameter in that case.

In the farther future we should allow users to omit that hyperparameter entirely as per #510

roystgnr:
truncation_error_precision is now properly disabled when we have a complete SVD basis, but when we have a truncated basis, truncation_error_precision is currently an improper prior placeholder.

truncation_error_precision needs to be user-disableable even in cases where we are truncating the SVD basis.  Unlike with observation_error_precision, disabling truncation_error_precision means fixing it at 0, not at 1.

roystgnr:
truncation_error_precision prior should be gamma(a_eta, b_eta) with default shape parameter 5, default scale parameter 200 - from the MATLAB code; this isn't in the paper.

brianw525:
Disabling truncation_error_precision means setting it to Inf and removing it from the MCMC.

PR #596, opened 2017-06-13
Use GammaRV prior for truncation_error_precision

roystgnr:
This gives us the default shape/scale values we want, and the options
infrastructure to change them, and sets up a proper gamma prior with
those values instead of an improper uniform prior.

This fixes more of the issues noted in #511

### DIFF
--- a/src/gp/inc/GPMSA.h
+++ b/src/gp/inc/GPMSA.h
@@ -382,6 +382,7 @@ public:
   typename ScopedPtr<VectorSpace<V, M> >::Type oneDSpace;
 
   // Truncation error precision
+  typename ScopedPtr<VectorSpace<V, M> >::Type truncationErrorPrecisionSpace;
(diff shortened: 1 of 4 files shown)


src/gp/inc/GPMSA.h
```diff
@@ -382,6 +382,7 @@ class GPMSAFactory
   typename ScopedPtr<VectorSpace<V, M> >::Type oneDSpace;
 
   // Truncation error precision
+  typename ScopedPtr<VectorSpace<V, M> >::Type truncationErrorPrecisionSpace;
```
briadam: This appears to get used without initialization.  Does default init suffice?

roystgnr: No, this is a bug.  Thanks!

roystgnr:
Should be okay now.

PR #602, opened 2017-06-14
GPMSA normalization options for output data

roystgnr:
This adds analogous scaling options to what we support for the input data, implements output autoscaling calculations within GPMSAOptions, and implements output scaling within GPMSA.

The "autoscale-all" regression test results changed, as expected, which unfortunately does not necessarily prove the results changed as expected.


src/gp/inc/GPMSAOptions.h
```diff
@@ -100,6 +100,12 @@ class GPMSAOptions
   void set_autoscale_minmax_scenario_parameter(unsigned int i);
 
   //! Do automatic normalization, using minimum and maximum values in
+  //  the supplied data, for output data i.
+  //
+  //  Normalized values will range from 0 to 1.
+  void set_autoscale_minmax_output(unsigned int i);
```
briadam: Good to have this for completeness, though Brian Williams recommends always using mean/variance scaling.

src/gp/inc/GPMSAOptions.h
```diff
@@ -119,6 +125,12 @@ class GPMSAOptions
   void set_autoscale_meanvar_scenario_parameter(unsigned int i);
 
   //! Do automatic normalization, using mean and variance of the
+  //  supplied data, for output i.
+  //
+  //  Normalized values will have mean 0 and standard deviation 1.
+  void set_autoscale_meanvar_output(unsigned int i);
```
briadam: As we think forward to the functional case, probably don't need scaling on a per-component basis.  More likely on a per-group basis, but having this generality is fine.

src/gp/src/GPMSA.C
```diff
@@ -1155,6 +1160,8 @@ GPMSAFactory<V, M>::setUpHyperpriors()
       V y_temp(Wy * y);
       y_temp -= Wy * B * yhat;
 
+      // At this point everything has already been normalized, so no
+      // more normalization to do?
```
briadam: I'm a little surprised that we don't have to scale the experiment covariance information provided by the user, as we did in examples/gp/linear_verif/  Or maybe you are with Wy above...

src/gp/src/GPMSAOptions.C
```diff
@@ -558,6 +594,30 @@ GPMSAOptions::set_final_scaling
...
+      for (unsigned int i=1; i < m_simulationOutputs.size(); ++i)
+        min_max_update(minOutput, maxOutput,
+                       (*m_simulationOutputs[i]));
+
+      for (unsigned int p=0; p != dimOutput; ++p)
+        if (m_autoscaleMinMaxAll ||
+            m_autoscaleMinMaxOutput.count(p))
+          {
+            if ((m_outputScaleMin.size() > p) &&
+                ((m_outputScaleMin[p] != 0) ||
```
briadam: Potentially dangerous test for floating point equality.

src/gp/src/GPMSAOptions.C
```diff
@@ -624,6 +684,36 @@ GPMSAOptions::set_final_scaling
...
+                        *m_simulationOutputs[i]);
+
+      varOutput /= n;
+
+      for (unsigned int p=0; p != dimOutput; ++p)
+        if (m_autoscaleMeanVarAll ||
+            m_autoscaleMeanVarOutput.count(p))
+          {
+            if ((m_outputScaleMin.size() > p) &&
+                ((m_outputScaleMin[p] != 0) ||
```
briadam: Potentially dangerous test for floating point equality.

src/gp/src/GPMSA.C
```diff
@@ -766,7 +767,8 @@ GPMSAFactory<V, M>::setUpEmulator()
 
   for (unsigned int i=0; i != m_numSimulations; ++i)
     for (unsigned int j=0; j != numOutputs; ++j)
-      (*simulationOutputMeans)[j] += (*m_simulationOutputs[i])[j];
+      (*simulationOutputMeans)[j] +=
```
briadam: Verify that this isn't doubly applying the mean shift between the code here and in normalized_output.

src/gp/src/GPMSA.C
```diff
@@ -776,7 +778,8 @@ GPMSAFactory<V, M>::setUpEmulator()
   for (unsigned int i=0; i != m_numSimulations; ++i)
     for (unsigned int j=0; j != numOutputs; ++j)
       simulation_matrix(i,j) =
-        (*m_simulationOutputs[i])[j] - (*simulationOutputMeans)[j];
+        this->m_opts->normalized_output(j, (*m_simulationOutputs[i])[j]) -
+        (*simulationOutputMeans)[j];
```
briadam: Verify that this isn't doubly applying the mean shift between the code here and in normalized_output.  (And in all places normalized_output is being called below.)

src/gp/src/GPMSA.C
```diff
@@ -904,7 +907,9 @@ GPMSAFactory<V, M>::setUpEmulator()
               // No fancy perturbation here
               unsigned int j = ex*numOutputs+outj;
 
-              Wy(i,j) = W_i(outi,outj);
+              Wy(i,j) = W_i(outi,outj) *
```
briadam: Perhaps this is accounting for the need to scale the user's observation covariance?

briadam (changes requested):
Change might not strictly be needed, but comments should be reviewed before merge.

src/gp/inc/GPMSAOptions.h
roystgnr: Yeah; I think when we add the partitioning capability, this won't be "output i" but rather "output subpartition i".

src/gp/src/GPMSA.C
roystgnr: This code is correct, but it's subtle.  If you're doing mean/variance normalization on a single index i, then simulationOutputMeans will be 0, and calculating it is a waste of time, but still gives the correct result.

But, if you're normalizing based on min/max, or a user-submitted range, or if index i is only one index of many in a partition, then the mean of the normalized values will not be zero, and yet we'll still want to subtract it off before doing the SVD.

src/gp/src/GPMSA.C
roystgnr: This was a bug, fixed in the latest commit; thanks.

src/gp/src/GPMSA.C
roystgnr: I was scaling it, but in the *wrong* direction, before that fix.

src/gp/src/GPMSAOptions.C
roystgnr: Nah.  The fill in set_output_scaling is with *exactly* 0 and 1, so we can rely on an exact test to tell us if something other than a fill happened.

Now, there is a loophole here: if someone manually sets 0 and 1, then requests automatic scaling, then we won't catch that and scream at them - by doing something insane *twice* they'll have evaded our sanity check...

briadam (approved):

PR #605, opened 2017-06-15
Deprecated *Options class cleanup

briadam:
In preparation for more flexible Options construct/parse/update APIs
(following example in GPMSAOptions), remove unused (and two years
deprecated) Options classes: EnvironmentOptions, MonteCarloSGOptions,
StatisticalForwardProblemOptions, StatisticalInverseProblemOptions

Didn't rename the header files to preserve backward-compatibility.

If you'd rather wait to merge  until the new options constructing / parsing is in place, that's fine with me.


briadam:
A related question is how long the project will maintain support for boost::program_options-based parsing.  Should new code support both, or does it suffice to just support getpot in new code?

dmcdougall:
That felt good.

briadam:
Great.  Any thoughts on my question about GetPot vs. Boost?

briadam:
Damon and Roy confirmed that since GetPot was the default for last release, only need to support that in new / modified code.  BPO code can be removed in these classes and likely more widely.

dmcdougall:
I was wrong.  We don't currently have a release where Boost program options is opt-in.  The commit to make it opt-in happened after the v0.57.0 release.

dmcdougall:
https://github.com/libqueso/queso/blob/dev/CHANGES#L6

dmcdougall:
I'm happy for you to remove the old *Options _classes_, as those have been deprecated for a while.  Boost parsing will have to stay until after we release 0.58.0.

Sorry for the confusion.

PR #607, opened 2017-06-20
Adding large GPMSA verification example

dmcdougall:
Just the scalar case for now.  To come are the multivariate and functional cases.

This needs #606 before the log priors will match.  Log likelihoods aren't matching yet, and we're aware we need to do some debugging to get to the bottom of it.


dmcdougall:
Rebased this on dev after merging #606.  Log priors should match.  Still looking at log likelihoods.

PR #608, opened 2017-06-20
Add cholSolve method to GslMatrix

### LINKED ISSUE #604: Implement a `choSolve` for `GslMatrix`
Useful for using the Cholesky decomposition computed by `chol()`.  Currently neither `invertMultiply` nor `svdSolve` use the result from `chol()`.

dmcdougall:
Fixes #604.

### DIFF
--- a/src/core/inc/GslMatrix.h
+++ b/src/core/inc/GslMatrix.h
@@ -192,7 +192,16 @@ public:
   //! This function solves the system A x = b using the singular value decomposition (U, S, V) of A which must have been computed previously with GslMatrix::svd (x=solMat, b=rhsMat).
   int               svdSolve                  (const GslMatrix& rhsMat, GslMatrix& solMat) const;
 
-
+  //! This function solves the system A x = b using a Cholesky decomposition of
+  //! A.  A is \c this, x is \c sol and b is \c rhs.
+  /*!
+   * If a Cholesky decomposition hasn't been cached internally, one is computed
+   * and cached for you.  Subsequent solves using the same matrix will then not
+   * re-compute the Cholesky factorisation but use the cached factorisation.
+   *
+   * Modifications to \c this will delete the cached Cholesky factorisation.
+   */
+  void cholSolve(const GslVector & rhs, GslVector & sol) const;
 
   //! This function multiplies \c this matrix by vector \c x and returns the resulting vector.
   GslVector  multiply                  (const GslVector& x) const;
@@ -410,6 +419,9 @@ private:
 
   //! GSL matrix for the LU decomposition of m_mat.
   mutable gsl_matrix*       m_LU;
+
+  //! GSL matrix for the LU decomposition of m_mat.
+  mutable gsl_matrix *      m_chol;
 
   //! Inverse matrix of \c this.
   mutable GslMatrix* m_inverse;
--- a/src/core/inc/GslMatrix.h
+++ b/src/core/inc/GslMatrix.h
@@ -424,7 +424,7 @@ private:
   //! GSL matrix for the LU decomposition of m_mat.
   mutable gsl_matrix*       m_LU;
 
-  //! GSL matrix for the LU decomposition of m_mat.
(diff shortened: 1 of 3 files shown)


briadam:
I reviewed. Just need to get to a full size keyboard. Mostly comment and error message changes, and some feedback for a later time about generalizing the class members.

src/core/inc/GslMatrix.h
```diff
@@ -192,7 +192,16 @@ class GslMatrix : public Matrix
   //! This function solves the system A x = b using the singular value decomposition (U, S, V) of A which must have been computed previously with GslMatrix::svd (x=solMat, b=rhsMat).
   int               svdSolve                  (const GslMatrix& rhsMat, GslMatrix& solMat) const;
 
-
+  //! This function solves the system A x = b using a Cholesky decomposition of
+  //! A.  A is \c this, x is \c sol and b is \c rhs.
```
briadam: Consider documenting that A should be a symmetric (or Hermitian if GSL supports complex) positive definite matrix.

src/core/src/GslMatrix.C
```diff
@@ -432,6 +441,50 @@ GslMatrix::chol()
   return iRC;
 }
 
+void
+GslMatrix::cholSolve(const GslVector & rhs, GslVector & sol) const
+{
+  queso_require_equal_to_msg(this->numCols(), rhs.sizeLocal(), "matrix and rhs have incompatible sizes");
+  queso_require_equal_to_msg(sol.sizeLocal(), rhs.sizeLocal(), "solution and rhs have incompatible sizes");
```
briadam, line 445: Since the solution vector sol has to be pre-sized prior to call, consider documenting.

src/core/src/GslMatrix.C
```diff
@@ -232,6 +237,10 @@ GslMatrix::copy(const GslMatrix& src)
 void
 GslMatrix::resetLU()
```
briadam: This function is unfortunately named, now that you support several matrix decompositions.  Since it's private, could consider a more general name, though chances of confusion are low.

src/core/inc/GslMatrix.h
```diff
@@ -411,6 +420,9 @@ GslVector& x) internally.*/
   //! GSL matrix for the LU decomposition of m_mat.
   mutable gsl_matrix*       m_LU;
 
+  //! GSL matrix for the LU decomposition of m_mat.
```
briadam: "for the LL^T decomposition"

src/core/src/GslMatrix.C
```diff
@@ -432,6 +441,50 @@ GslMatrix::chol()
...
+    m_chol = gsl_matrix_calloc(this->numRowsLocal(), this->numCols());
+    queso_require_msg(m_chol, "gsl_matrix_calloc() failed");
+
+    iRC = gsl_matrix_memcpy(m_chol, m_mat);
+    queso_require_msg(!(iRC), "gsl_matrix_memcpy() failed");
+
+    gsl_error_handler_t * oldHandler;
+    oldHandler = gsl_set_error_handler_off();
+
+    iRC = gsl_linalg_cholesky_decomp(m_chol);
```
briadam: The code here could be shared with the chol() function, though the behavior and error handling are slightly different.

briadam (approved):
Overall looks good.  Merge would be fine; comments are offered for consideration.

src/core/src/GslMatrix.C
dmcdougall: Yeah, I didn't want to change behaviour of chol() as it doesn't throw any exceptions.  You're right though.

Also, speaking of exceptions, we should be putting `m_chol` and the like in `ScopedPtr`s.

dmcdougall: I settled for a `SharedPtr` because `ScopedPtr`s don't support custom deleters.

briadam:
Only did a quick review.  Looks good from what I saw.

PR #613, opened 2017-07-26
Fix a couple of bugs in the GPMSA implementation

dmcdougall:
1.  We were multiplying by the observation precision and not dividing by it
2.  There was an indexing error in getting the emulator data precision variable
3.  There was an error in computing the sample variance (that was then used for normalising the simulation and experimental output to N(0, 1))


src/gp/src/GPMSA.C
```diff
@@ -363,7 +363,10 @@ GPMSAEmulator<V, M>::lnValue(const V & domainVector,
     // Add small white noise component to diagonal to make stuff +ve def
     // = "small ridge"
     // Barely alluded to, never described, in Higdon et. al.
-    const double emulator_data_precision = domainVector[dimSum-1-(numOutputs>1)];
+    const double emulator_data_precision =
+      m_opts.m_calibrateObservationalPrecision ? domainVector[dimSum-2-(numOutputs>1)]
+                                               : domainVector[dimSum-1-(numOutputs>1)];
+
```
roystgnr: Can we just subtract the bool here (as is done in GPMSAOptions with the same option, or here with (numOutputs>1)) rather than using a ternary operator?

dmcdougall: Sure, update incoming.

dmcdougall:
These broke the gpmsa regression test, obviously.  I've updated the regression solution.

roystgnr:
Looks good (except insofar as it's quite embarrassing).

dmcdougall:
Don't beat yourself up.  We would have caught it earlier if our regression test were actually a unit test (coming soon).

PR #580, opened 2017-05-05
GPMSA Example: Brian Williams' linear verification

briadam:
@roystgnr @dmcdougall @brianw525 This starts to address #579 by implementing Brian's test problem in a QUESO example.  It runs and doesn't crash (issues a bunch of warnings from one particular spot though.)

In terms of your development workflow, I wasn't sure if you want this to be managed on a branch until the test is verified vs. merged to devel. Feel free to guide me as to how to work.

Add initial version of Bayes linear verification problem due to Brian
Williams

[1] Adams, B.M., Coleman, Kayla, Hooper, Russell W., Khuwaileh, Bassam
A., Lewis, Allison, Smith, Ralph C., Swiler, Laura P., Turinsky, Paul
J., Williams, Brian J., "User Guidelines and Best Practices for CASL
VUQ Analysis Using Dakota, Sandia Technical Report 2016-11614 (also
CASL-U-2016-1233-000), September 2016.

PR #615, opened 2017-07-27
Adding large GPMSA verification example

dmcdougall:
Replaces #607.


examples/gp/linear_verif/gpmsa_linear_verif_pdf.C
```diff
@@ -0,0 +1,614 @@
...
+  }
+
+  // Experiment observation error (also needs to be scaled by
+  // simulation variance)
+  for (unsigned int i = 0; i < numExperiments; i++)
+    (*experimentMat)(i, i) = std::pow(0.05, 2.0);
+
+  // Read in data and store the standard deviation of the simulation
+  // data (ignored for now).
+  readData("sim_scalar.dat", "y_exp_scalar.txt", simulationScenarios, paramVecs,
```
roystgnr: Are these files included in the PR anywhere?

roystgnr: Wait, never mind, those are the files that were merged in #580.

dmcdougall:
@roystgnr Ok I think I'm done here too.  Let me know if you have any comments.

dmcdougall:
@briadam This is the pull request where I destroyed all your hard work.  I'll resurrect it.

briadam:
Hardly destroyed... just morphed it into a necessary prerequisite test. ;-)  

Honestly, you'll probably do better to just write an inverse problem departing from scalar_pdf_large without looking at what I did as you'll likely then immediately see the issues I ran into with the initial positions and proposal covariance.  I can see having to scale the proposal covariance as we discussed today, but it's possible at this point (given all the work you've done) that if you take the initial point at the mean and proposal covariance to be the prior variance, you may fare better than I did, and it'll just work.

dmcdougall:
Well, it'll be a pull request in a couple of minutes and then you can tell me what to do with it :)

PR #616, opened 2017-07-31
Normalization changes

roystgnr:
Autoscale by simulation data only, to better match the MATLAB code.

Gracefully handle the "we have nothing to scale by if we leave out the experimental data" cases; even with a dummy variable where we don't care about the scaling, NaNs spread like weeds.

### DIFF
--- a/src/gp/src/GPMSAOptions.C
+++ b/src/gp/src/GPMSAOptions.C
@@ -61,7 +61,8 @@
 namespace { // Anonymous namespace for helper functions
 
 template <typename V>
-void min_max_update(V & min, V & max, const V & new_data)
+void min_max_update(V & min, V & max, const V & new_data,
+                    const char * warn_on_update = NULL)
 {
   unsigned int dim = min.sizeGlobal();
   queso_assert_equal_to(dim, max.sizeGlobal());
@@ -69,6 +70,19 @@ void min_max_update(V & min, V & max, const V & new_data)
 
   for (unsigned int p=0; p != dim; ++p)
     {
+      if (warn_on_update)
+        {
+          if (min[p] > new_data[p])
+            queso_warning("Experimental " << warn_on_update << " " <<
+                          new_data[p] << " at index " << p <<
+                          " is below minimum simulation " <<
+                          warn_on_update << " " << min[p]);
+          if (max[p] < new_data[p])
+            queso_warning("Experimental " << warn_on_update << " " <<
+                          new_data[p] << " at index " << p <<
+                          " is above maximum simulation " <<
+                          warn_on_update << " " << max[p]);
+        }
       min[p] = std::min(min[p], new_data[p]);
       max[p] = std::max(max[p], new_data[p]);
     }
@@ -550,16 +564,18 @@ GPMSAOptions::set_final_scaling
 
       V minScenario(*m_simulationScenarios[0]);
 
+      // Only use simulation data for scaling.
       for (unsigned int i=1; i < m_simulationScenarios.size(); ++i)
         min_max_update(minScenario, maxScenario,
                        (*m_simulationScenarios[i]));
 
-      // To match Matlab code, only use simulation data for scaling.
-/*
+      // But check the experimental data, and yell at the user if it
+      // falls outside the simulation data range, because they
+      // probably don't want to be extrapolating with their GP.
       for (unsigned int i=0; i < m_experimentScenarios.size(); ++i)
         min_max_update(minScenario, maxScenario,
-                       (*m_experimentScenarios[i]));
-*/
+                       (*m_experimentScenarios[i]),
+                       /*warn_on_update =*/ "scenario parameter");
 
       for (unsigned int p=0; p != dimScenario; ++p)
         if (m_autoscaleMinMaxAll ||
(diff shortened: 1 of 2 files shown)


roystgnr:
Pretty sure this PR doesn't *actually* break test_uqGaussianVectorRVClass on clang.

Stupid stochastic tests.  Are we not using a fixed PRNG seed in the test suite?  That wouldn't fix "new algorithm calls for random numbers in a different order" false positives, but it would at least fix "old algorithm gets different random numbers this time" false positives.

src/gp/src/GPMSAOptions.C
```diff
@@ -554,9 +554,12 @@ GPMSAOptions::set_final_scaling
         min_max_update(minScenario, maxScenario,
                        (*m_simulationScenarios[i]));
 
+      // To match Matlab code, only use simulation data for scaling.
```
dmcdougall: Just delete this code instead of commenting it out.

src/gp/src/GPMSAOptions.C
```diff
@@ -634,9 +637,12 @@ GPMSAOptions::set_final_scaling
         mean_var_update(n, meanScenario, varScenario,
                         *m_simulationScenarios[i]);
 
+      // To match Matlab code, only use simulation data for scaling.
```
dmcdougall: Ditto.

dmcdougall (changes requested):

src/gp/src/GPMSAOptions.C
roystgnr: I failed to come up with a decent middle ground there, huh?  I was trying to avoid both the confusing Scylla of "just delete the code, and leave future generations with no indication of why we're ignoring half of our input data" and the passive-aggressive Charybdis of "// Only use simulation data for scaling.  This is necessary to match Matlab code output, not for any sane reason."  ;-)

Would it be reasonable for me to leave the code in, but wrapped in an "autoscale_including_experiment_data" option (which defaults to false, if you like)?  Or should I move the documentation to the autoscale method declarations?  Or both?  (I'm now leaning toward "both".)

I'm seriously insisting on *some* sort of documentation to exist regarding why we're making this choice.  "Why are we normalizing numbers to a (min,max) range which will leave numbers outside that range?" is a question with a defensible answer, but "Why are we doing so without telling users?" is not.

brianw525:
The Gaussian process surrogate used for calibration in GPMSA is built using the simulation data only. Therefore, it is really only valid on the convex hull of the simulation inputs (scenario and calibration). Anything outside of that is an extrapolation, and we should never extrapolate with surrogates as a general rule. The choice of simulation inputs only for normalization is therefore motivated by the fact that they constitute the training data for the surrogate and the interpretation of correlation length parameters is based on the design region chosen for generating the simulation data.

No user should be calibrating to experimental scenarios that fall outside of the range of scenarios determined by the simulation data. Since we are normalizing with respect to the simulation inputs, I would suggest that QUESO (or Dakota) throws an error (or warning) if any of the experimental scenario variables take a normalized value greater than 1 or less than 0. The practitioner should ensure that the range of scenarios passed to the simulator covers the range of experimental scenarios used to calibrate the model parameters.

roystgnr:
We have more than one gaussian process in play here, though, right?  The GP modeling the simulator is calibrated using only simulator data, but the GP modeling the discrepancy must be calibrated using experiment data points.  Is it really obvious that normalizing the former to a too-wide data range would be worse than normalizing the latter to a too-narrow range?

I'll put this in docs, regardless.  I like your idea about a warning, too, and I can add a bool+getter in case Dakota wants to turn that warning into an error.

brianw525:
Yes we do, however both the GP modeling the simulator AND the GP modeling the discrepancy are involved in the calibration. Therefore if we have experimental scenarios outside the range of the simulator scenarios, we are extrapolating regardless of what normalization we use and this should not be allowed as a matter of principle.

Bottom line is we need to be normalizing scenarios (and calibration parameters of course) with respect to the simulator design only. This is the convention utilized in the Matlab code and there is no justified reason for changing it. It needs to be done the same way in the Matlab code and in QUESO to make verification feasible.

src/gp/src/GPMSAOptions.C
```diff
@@ -550,16 +564,18 @@ GPMSAOptions::set_final_scaling
...
-/*
+      // But check the experimental data, and yell at the user if it
+      // falls outside the simulation data range, because they
+      // probably don't want to be extrapolating with their GP.
       for (unsigned int i=0; i < m_experimentScenarios.size(); ++i)
         min_max_update(minScenario, maxScenario,
-                       (*m_experimentScenarios[i]));
-*/
+                       (*m_experimentScenarios[i]),
+                       /*warn_on_update =*/ "scenario parameter");
```
dmcdougall: If `min_max_update` is still being called on the experiment scenarios, then isn't the min and max including the experiment scenarios in the normalization?

dmcdougall: If min_max_update is still being called on the experiment scenarios, then isn't the min and max including the experiment scenarios in the normalization?

dmcdougall (changes requested):

src/gp/src/GPMSAOptions.C
dmcdougall: Sorry, that comment is horribly worded.  Let me try again.

If `min_max_update` is still being called on the experiment scenarios, then we're including the experiment scenarios in the normalization.

roystgnr: Yeah, sorry; that last commit should fix it.

dmcdougall (approved):

PR #618, opened 2017-08-02
Fix hdf5 m4

### LINKED ISSUE #610: hdf5.m4 doesn't snarf $HDF5_DIR correctly
```bash
$ ./configure --with-hdf5=/path/to/hdf5
```

correctly defines `HAVE_HDF5` in `config_queso.h` but

```bash
$ export HDF5_DIR=/path/to/hdf5
$ ./configure
```

does not define `HAVE_HDF5`.

dmcdougall:
Fixes #610.


roystgnr:
Could you push these changes to pecos/common svn too?  I know nobody uses that directly anymore, but so long as everyone's too lazy to give it its own git repo, that's the closest thing to an "authoritative" fork of these utilities that we've got.

PR #621, opened 2017-08-21
Fix parsing for max SVD basis vectors input option

dmcdougall:
We had an input file option for the maximum number of SVD basis vector to
use for the emulator, but we never parsed it.  Now we parse it.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -5,6 +5,7 @@ Simulation, and Optimization.
 
 Version 0.58.0 (Date?)
   * Make Boost program_options opt-in
+  * Multivariate GPMSA bug fixes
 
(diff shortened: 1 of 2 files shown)


dmcdougall:
No regression test solution update needed.

CHANGES
```diff
@@ -5,6 +5,7 @@ Simulation, and Optimization.
 
 Version 0.58.0 (Date?)
   * Make Boost program_options opt-in
+  * Multivariate GPMSA bug fixes
```
roystgnr, line 8: Fine, rub it in.

roystgnr:
Thanks!

PR #622, opened 2017-08-21
Fix standard basis calculation for discrepancy matrices

dmcdougall:


dmcdougall:
@roystgnr This relies on a commit cherry-picked from your `gpmsa_discrepancy_fix` branch.

roystgnr:
First thought: wait, I have a gpmsa_discrepancy_fix branch?

Upon examination of git log: I actually split that bugfix out from gpmsa_new_functional into its own branch, then FORGOT TO SUBMIT A PR FROM THAT BRANCH.  I'm not sure whether that's more or less embarrassing than not splitting out the bugfix properly at all.

PR #623, opened 2017-08-21
Fix gpmsa svd order

dmcdougall:
Please merge #622 before merging this.


roystgnr:
Hmm... this one actually needs a couple things before merge:

1) Github is still showing me the full diff-from-branch-point; I'd like to wait until tomorrow morning to hit the command line and make sure that the diff-from-master now only includes the last commit, as expected.

2) Depending on GSL ordering conventions, when I've already gotten GSL ordering conventions wrong, scares me.  We should add some kind of queso_assert_less_equal(next_singular_value, last_singular_value) for safety's sake, especially if there's a chance of us moving away from GSL eigensolves in the future.

dmcdougall:
> I'd like to wait until tomorrow morning to hit the command line and make sure that the diff-from-master...

diff-from-dev?

> We should add some kind of queso_assert_less_equal...

No need, [we explicitly put them in ascending order](https://github.com/libqueso/queso/blob/dev/src/core/src/GslMatrix.C#L1487)

> ...for safety's sake, especially if there's a chance of us moving away from GSL eigensolves in the future

Now, that I agree with.

dmcdougall:
Done.

roystgnr (approved):
Thanks for the extra error checking; :+1: once Travis is happy.

PR #624, opened 2017-08-22
Fix W_y precision calculation

dmcdougall:
W_y is a precision matrix, not a covariance matrix.

Please merge #623 before merging this.  Also, this PR isn't complete.  There are *unit* (not regression) tests that fail, and I haven't figured out why yet.  I also need to add commits that:

1.  update the internal name of these matrices from the `*covariance*` flavour to the `*precision*` flavour;
2.  update the comments in the same manner.


dmcdougall:
Oh I see why the unit tests are failing now.

roystgnr:
Didn't we decide that we need to rename m_observationErrorMatrix, now that W_y is actually holding precisions rather than errors?  I presume this has something to do with the scalar unit test breakage, because we actually are using that error matrix as an error matrix in that case?

dmcdougall:
Ok I've fixed the unit test failures.

I'd rather keep the m_emulatorErrorMatrix verbage around, because when we get to lnValue, the error matrices are better to work with than the precision matrices.  Check out the commit for the solution I implemented.

src/gp/src/GPMSA.C
```diff
@@ -968,6 +972,9 @@ GPMSAFactory<V, M>::setUpEmulator()
     (*BT_Wy_B_inv)(i,i) +=
       this->m_opts->m_observationalCovarianceRidge;
 
+  // Emulator expects the error matrix, so invert Wy to get it.
+  m_observationErrorMatrix.reset(new M(Wy.inverse()));
+
```
roystgnr: This is a block diagonal matrix where the blocks are what the user provided.  Why are we first inverting the blocks, then making that big block diagonal matrix, then inverting the big matrix?

src/gp/src/GPMSA.C
```diff
@@ -894,25 +894,29 @@ GPMSAFactory<V, M>::setUpEmulator()
...
   const Map Wy_row_map(Wyrows, 0, comm);
 
-  m_observationErrorMatrix.reset
-    (new M(env, Wy_row_map, Wyrows));
-
   M& B = *m_BMatrix;
-  M& Wy = *m_observationErrorMatrix;
+
+  // Observation precision matrix
+  M Wy(env, Wy_row_map, Wyrows);
```
roystgnr, line 900: I guess it's okay for Wy to be a local matrix now because we only use it indirectly via BT_Wy_Binv?

dmcdougall: Yep.

src/gp/src/GPMSA.C
dmcdougall: Because I'm lazy.  I'll update it to be less stupidly inefficient.

roystgnr: Thanks!  The inefficiency doesn't bother me too too much (since it's once per run, not once per sample), but you had me questioning my sanity.  "Is there something wrong with my basic understanding of linear algebra identities?  Is the inverse-of-the-inverse somehow better for some numerical reason?"

dmcdougall: No.  In fact the inverse-of-the-inverse is probably worse in terms of numerical accuracy?  ¯\\\_(ツ)\_/¯

src/gp/src/GPMSA.C
```diff
@@ -899,13 +899,17 @@ GPMSAFactory<V, M>::setUpEmulator()
   // Observation precision matrix
   M Wy(env, Wy_row_map, Wyrows);
 
+  m_observationErrorMatrix.reset(new M(env, Wy_row_map, Wyrows));
+
   for (unsigned int ex = 0; ex != m_numExperiments; ++ex)
     {
       const M & D_i = m_discrepancyMatrices[ex];
 
+      const M & Sigma_i = *m_observationErrorMatrices[ex];
```
roystgnr: You've got the handy readably-named reference here, why not use it when taking the inverse below?

roystgnr:
:+1: once Travis is happy.

src/gp/src/GPMSA.C

PR #625, opened 2017-08-22
Fix an indexing in error in BT_Wy_B ridge addition

dmcdougall:
We were indexing over the number of rows of B, but this is wrong.  The
number of rows it has is the number of columns of B.

### DIFF
--- a/src/gp/src/GPMSA.C
+++ b/src/gp/src/GPMSA.C
@@ -962,14 +962,18 @@ GPMSAFactory<V, M>::setUpEmulator()
   // overridden by user options.
   //
   // This is "DKridge" in the MATLAB prototype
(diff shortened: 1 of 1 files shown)


roystgnr:
Did we have Brows==Bcols before?  I don't see how this bug caused neither solution changes nor segfaults otherwise.

dmcdougall:
I guess we must have.

roystgnr:
Could we throw two asserts in there, demanding that BT_Wy_B_inv has the same number of rows and cols and that they're both Bcols?

roystgnr:
I know, closing the barn door after the horses have left, but we might refactor at some point.

dmcdougall:
Done.

roystgnr (approved):
Looks good, ready to merge when Travis is.

PR #628, opened 2017-08-27
Fix indexing of emulator precisision in setup

dmcdougall:
We don't need to conditionally check if `numOutputs > 1`.  The number of
emulator precision variables is always `num_svd_terms`, and `num_svd_terms`
is `numOutputs` by default, so in the scalar case it will be 1.

This commit also updates the regression test solutions.


dmcdougall:
@roystgnr In #511 you said this:

> When we're in a multivariate or functional case with a non-truncated SVD, truncation_error_precision (lambda_eta in the paper's multivariate case) needs to go unused!

I believe this is where the `numOutputs > 1` confusion came from.  I could be wrong, though.

roystgnr:
Looks good but I want to take a second look in the morning.

I was creating too-large a (hyper)parameter space, so no OOB vector accesses were being detected?

dmcdougall:
> I was creating too-large a (hyper)parameter space, so no OOB vector accesses were being detected?

I guess so?

We should perhaps be better about running our tests in debug mode; [we turn off GSL's range checking when we aren't in debug mode](https://github.com/libqueso/queso/blob/dev/src/core/inc/Defines.h#L57-L60).

roystgnr:
I run everything in dbg mode, but GSL's checks don't help when our vector is too long or mis-indexed internally, only if it's too long or mis-indexed externally.

test/test_gpmsa/test_gpmsa_vector.C
```diff
@@ -343,17 +342,16 @@ int main(int argc, char ** argv) {
...
-  proposalCovMatrix(9, 9)   = 7.6032e-04 / scale;  // emulator corr str
-  proposalCovMatrix(10, 10) = 8.3815e-04 / scale;  // emulator corr str
-  proposalCovMatrix(11, 11) = 7.5412e-04 / scale;  // emulator corr str
-  proposalCovMatrix(12, 12) = 0.2682 / scale;  // emulator corr str
-  proposalCovMatrix(13, 13) = 0.0572 / scale;  // emulator corr str
-  proposalCovMatrix(14, 14) = 1.3417 / scale;  // discrepancy precision
-  proposalCovMatrix(15, 15) = 0.3461 / scale;  // discrepancy corr str
-  proposalCovMatrix(16, 16) = 495.3 / scale;  // emulator data precision
-  proposalCovMatrix(17, 17) = 0.4953 / scale;  // observation error precision
+  proposalCovMatrix(7, 7)   = 0.6058 / scale;  // emulator corr str
```
roystgnr, line 345: As long as we're fixing indexing... we should fix these comments later.  Truncation error precision should be off for this run, and we have 2 outputs, in which case we should follow the 5 non-hyper parameters with 2 emulator precision hyperparameters.  The "emulator precision" vs "weights precision" was a bit of confusion months ago on my part.

1 scenario and 5 uncertain parameters means 6 correlation strengths, and everything after that follows correctly too.

PR #629, opened 2017-08-28
Fix gpmsa discrepancy precision indexing

dmcdougall:
Please merge #628 before merging this PR.

PR #630, opened 2017-08-28
Fix gpmsa discrepancy corr str indexing

dmcdougall:
Please merge #629 (and its dependencies) before merging this PR.


src/gp/src/GPMSA.C
```diff
@@ -1453,7 +1453,7 @@ GPMSAFactory<V, M>::setUpHyperpriors()
     (new VectorSpace<V, M>
      (this->m_env,
       "",
-      dimScenario,
+      numOutputs*dimScenario,
```
roystgnr: Could we get "F = numOutputs" into the code, not just in the comments?  Even if it's just "const unsigned int num_discrepancy_groups = numOutputs", and then we use num_discrepancy_groups here and below, that would be very helpful when I make F multivariable-vs-functional later.

PR #631, opened 2017-08-28
Modify logic for truncation error precision calibration

dmcdougall:
We don't compare `num_svd_terms` against `numOutputs` any more.  Instead we compare against the number of nonzero eigenvalues (up to some tolerance).

### DIFF
--- a/src/gp/src/GPMSA.C
+++ b/src/gp/src/GPMSA.C
@@ -76,13 +76,6 @@ GPMSAEmulator<V, M>::GPMSAEmulator(
 
   queso_assert_equal_to
     (m_simulationOutputs[0]->map().Comm().NumProc(), 1);
-
-  const unsigned int numOutputs =
-    this->m_experimentOutputSpace.dimLocal();
-
-  this->num_svd_terms = this->m_opts.m_maxEmulatorBasisVectors ?
-    std::min((unsigned int)(this->m_opts.m_maxEmulatorBasisVectors), numOutputs) :
-    numOutputs;
 }
 
 template <class V, class M>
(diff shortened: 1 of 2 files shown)


src/gp/src/GPMSA.C
```diff
@@ -847,6 +837,30 @@ GPMSAFactory<V, M>::setUpEmulator()
     queso_assert_less_equal(SM_singularValues[i], SM_singularValues[i+1]);
   }
 
+  // Count the number of "nonzero" eigenvalues (eigenvalues over a tolerance)
+  double eigenvalue_tolerance = 1e-10;
+  for (unsigned int i = 0; i < SM_singularValues.sizeLocal(); i++) {
+    // All eigenvalues should be positive, so we don't take fabs
+    if (SM_singularValues[i] > eigenvalue_tolerance) {
```
roystgnr, line 844: That tolerance ought to be relative, not absolute... but I guess we're rescaling data anyway so it doesn't matter much, and we're going to turn this into a user-controllable knob eventually so we can tweak it then.

src/gp/src/GPMSA.C
```diff
@@ -1034,6 +1048,10 @@ GPMSAFactory<V, M>::setUpEmulator()
       *this->BT_Wy_B_inv,
       *this->KT_K_inv,
       *this->m_opts));
+
+  // FIXME: epic hack.  pls fix.
```
roystgnr, line 1052: This isn't awful.  There's not really a perfect solution to the dichotomy of "put arguments in the constructor so you can enforce consistency and never leave objects in an invalid state" vs "put arguments in setter functions so you don't have to change the constructor every freaking time you add a new option".

roystgnr:
I'm pretty sure that in your second commit I'll get a merge conflict with Every. Single. Line.

Otherwise looks great.

PR #632, opened 2017-08-28
Fix yet another emulator precision indexing bug

dmcdougall:
Although this could have been avoided, please merge #631 before merging this PR.


src/gp/src/GPMSA.C
```diff
@@ -401,7 +401,7 @@ GPMSAEmulator<V, M>::lnValue(const V & domainVector,
           covMatrix(i,j) += BT_Wy_B_inv(i,j) * inv_lambda_y;
 
       const double emulator_precision =
-        domainVector[dimParameter+1];
+        domainVector[dimParameter+(num_svd_terms<num_nonzero_eigenvalues)];
```
roystgnr: I think this code is wrong, but I don't think this is the right fix here.  This is the lambda_eta term, isn't it?  So if num_svd_terms < num_nonzero_eigenvalues then we should be indexing with dimParameter; if not then we shouldn't be adding the inv(lambda_eta * K^T * K) term at all!

roystgnr (changes requested):

src/gp/src/GPMSA.C
dmcdougall: Oh, I see.  This is the truncation error precision, not the emulator precision.  In that case I'll change the name of that variable as well as it's misleading.

roystgnr: Please do, and sorry for the confusion.

roystgnr (approved):

roystgnr:
:+1: when Travis is happy

PR #633, opened 2017-08-31
Only add emulator data precision to emulator part

dmcdougall:
In the scalar case this'll boil down to nothing; that is we add the
inverse of the emulator data precision to every diagonal element.  In
the multivariate case we skip over the
num_emulator_bases*num_experiments piece that doesn't correspond to the
emulator part of the covariance matrix.

This commit also updates a regression test solution.  But not the other
one.  Not sure why.


src/gp/src/GPMSA.C
```diff
@@ -380,10 +380,11 @@ GPMSAEmulator<V, M>::lnValue(const V & domainVector,
...
-    for (unsigned int disc = 0; disc != num_discrepancy_bases;
-         ++disc)
-      covMatrix(disc*m_numExperiments+i,
-                disc*m_numExperiments+i) += nugget;
+    unsigned int discrepancy_basis_offset =
+      numOutputs == 1 ? 0 : num_discrepancy_bases;
+
+    covMatrix(discrepancy_basis_offset*m_numExperiments+i,
+              discrepancy_basis_offset*m_numExperiments+i) += nugget;
   }
```
roystgnr: Wait, this can't be right, can it?  This is supposed to be adding to a diagonal, or at least a diagonal of an on-diagonal block, not a *single* entry?  I thought you had just discussed changing the offset, not getting rid of the for loop.

roystgnr: Summarizing offline discussion: it looks like we want to *replace* the for loop; we should be looping over num_svd_terms here, not num_discrepancy_bases?

dmcdougall: Right.  I was about to summarise our in-person discussion but you nailed it.

roystgnr (approved):
Looks good now to me.

PR #634, opened 2017-09-13
Resurrect Brian's full GPMSA inverse problem

dmcdougall:
I didn't preserve the mean and variance calculation of the simulations because I figured, since we're doing this in QUESO, it didn't matter.


briadam (approved):
This can be merged as-is and then probably updated to comment out explicit defaults in the input file and remove hacks on initial and covariance values as we debug.

Once we get this in, I'll probably open a separate issue to achieve my original goal: to verify that we can run DRAM with mostly default GPMSA settings and get posteriors that are at least visually comparable to @brianw525 linear verification in Matlab.  The goal of that will be to get this example to the minimum set of non-default things to get reasonable behavior.

PR #635, opened 2017-09-14
Fix eta row/col major ordering

dmcdougall:
The (K^T K)^{-1} K^T matrix is expecting to act on a vector \eta such
that the simulations are contiguous in memory, not the outputs.


dmcdougall:
I probably ought to do this for the experiments, too.

roystgnr:
I was *just* thinking that "gee, it's been too long since I rebased three dozen functional branch commits"!

roystgnr:
Wait, is this right?  The paper says "We define the mn η vector η to be the concate-
nation of all m simulation output vectors η = vec()", where apparently Github renders a copied and pasted capital xi as a telephone for some reason.  (Upon edit: the telephone turns into a diamond-with-a-question-mark for me in the rendered markdown.  Not sure if that's an improvement.)

If we're concatenating simulation output vectors, then incrementing output index should add 1 to the concatenated index, because that just moves up one within the same subvector, and incrementing simulation index should skip to the next subvector, thereby adding numOutputs to the concatenated index.  That's what the original code does, and that's the opposite of what this PR does.  So what gives?

dmcdougall:
The paper also states the mn_η × mp_η matrix K is given by K = [I_m ⊗ k_1; ...; I_m ⊗ k_{p_η} ], and that K_η = [k_1 ; ...; k_{p_η} ].  So, I'll answer your question with another question:  what does this notation mean?  Listing vectors separated by columns could mean "construct the matrix with those vectors as columns", or "construct the matrix with those vectors as rows".

I agree with your interpretation of the concatenation of simulations, but it really doesn't matter.  Whatever we choose has to be consistent with however K is constructed, since we need to compute K^T η, but as far as I can see the paper isn't clear about how that is done.

brianw525:
These calculations are made much clearer in the attached document, which I would recommend following as it is far more efficient than doing it the way stated in the paper.

dmcdougall:
@brianw525 I don't see an attachment (GitHub problem maybe).  If you're talking about the helpful writeup you did for the scalar and multivariate cases you emailed out, thank you for the reminder.  Indeed as you point out in the writeup:

1.1.2 Emulator
The reduced simulator data vectors w˜j corresponding to the j-th simulator run are components
of the \hat{z} vector from the Higdon et al. paper (elements (p_δ + p_η)n + j,(p_δ + p_η)n +
m + j, . . . ,(p_δ + p_η)n + (p+η − 1)m + j).

If I'm understanding this correctly, the outputs are contiguous in \hat{z}, not the simulations.  In other words, this pull request is correct and the implementation we had before this change was incorrect.

Someone correct me if I'm off base here.

brianw525:
Indeed that’s the write-up Damon. Now I know this doesn’t like attachments …

The w~j that you actually calculate are scattered in \hat{z} according to the indexing in the write-up which you stated. Recall \hat{z} = (v, u, w). The “w” component of \hat{z} is equal to vec(W) for the W matrix given in Section 1.1.2 of that write-up.

The essential point here is that you don’t have to actually calculate the K matrix of the paper. You just need K_\eta which we get from the SVD. This goes to my point about efficiency.

Section 1.1.1 of the same write up explains how you can get the “(v,u)” part of \hat{z} really efficiently as well. Again you can do it piecemeal and you don’t need to compute any “big” matrices. The v_i~ and u_i~ that you actually calculate are again scattered in \hat{z} but there is an easy recipe for converting these into the “(v,u)” of \hat{z} [namely, (vec(U), vec(V)) for the U and V matrices given in the write-up].

brianw525:
Oops the end of my last email should say (vec(V), vec(U)), sorry!

From: "Williams, Brian J" <brianw@lanl.gov>
Date: Friday, September 15, 2017 at 3:18 PM
To: libqueso/queso <reply@reply.github.com>
Subject: Re: [libqueso/queso] Fix eta row/col major ordering (#635)

Indeed that’s the write-up Damon. Now I know this doesn’t like attachments …

The w~j that you actually calculate are scattered in \hat{z} according to the indexing in the write-up which you stated. Recall \hat{z} = (v, u, w). The “w” component of \hat{z} is equal to vec(W) for the W matrix given in Section 1.1.2 of that write-up.

The essential point here is that you don’t have to actually calculate the K matrix of the paper. You just need K_\eta which we get from the SVD. This goes to my point about efficiency.

Section 1.1.1 of the same write up explains how you can get the “(v,u)” part of \hat{z} really efficiently as well. Again you can do it piecemeal and you don’t need to compute any “big” matrices. The v_i~ and u_i~ that you actually calculate are again scattered in \hat{z} but there is an easy recipe for converting these into the “(v,u)” of \hat{z} [namely, (vec(U), vec(V)) for the U and V matrices given in the write-up].

roystgnr:
> The reduced simulator data vectors w˜j corresponding to the j-th simulator run are components
of the \hat{z} vector from the Higdon et al. paper (elements (p_δ + p_η)n + j,(p_δ + p_η)n +
m + j, . . . ,(p_δ + p_η)n + (p+η − 1)m + j).
>
> If I'm understanding this correctly, the outputs are contiguous in \hat{z}, not the simulations. In other words, this pull request is correct and the implementation we had before this change was incorrect.

I'm still not seeing how this matches up to Higdon et. al.  "The concatenation of the simulation output vectors" clearly means "each simulation has an output vector, and you concatenate them". "vec(Sigma)" in Matlab-speak means "the matrix Sigma has columns, and you concatenate them", and Sigma is defined as an n_eta by m matrix: m columns, one for each simulation, with n_eta outputs each.

I agree that we need eta and K to be consistent here, but if we make them consistent with each other and inconsistent with the paper then we're just going to end up breaking something again when someone tries to use the paper as documentation.  Even if our contention is that the paper is wrong (is it?), we'd need a comment with big flaming caps saying "the paper is wrong"!

So let's look at K?

In the code we have K as a matrix with numSimulationOutputs * m_numSimulations rows (which is n_eta * m, matching the paper) and m_numSimulations * num_svd_terms columns (which is m * p_eta, matching the paper).  The svd output vectors are treated as column vectors, are repeated m times (as per the I_m tensor product in the paper, offset by n_eta rows and 1 column each time), and the matrices from each (I_m tensor k_i) are concatenated horizontally.  So each submatrix has m * n_eta rows and m columns, and the full matrix is of the right size.

Maybe that's what I'm doing wrong somehow?  The obvious alternative would be treating the svd output vectors as row vectors, and doing the submatrix concatenation vertically, but that doesn't work.  That's what I'd assumed originally, IIRC (since semicolon is matlab-speak for "concatenate vertically"), but that would give a final matrix size of n_eta * m columns and m * p_eta rows, not the other way around.

brianw525:
I’ve suggested a few times that we need to go with the calculation approach recommended in the document I sent you after our last Austin visit. It gives the correct answers exactly how the Matlab code computes them. Please do this.

When I have time it will not be too difficult to prove that what’s in the paper is equivalent to what is in this document. However, the end of the fiscal year is not the appropriate time to spend any effort on this.

dmcdougall:
Replaced by #639.

PR #636, opened 2017-09-14
Fix Wy matrix in residual calculation

dmcdougall:
Please merge #635 before merging this PR.


src/gp/src/GPMSA.C
```diff
@@ -1235,7 +1235,7 @@ GPMSAFactory<V, M>::setUpHyperpriors()
       }
 
       M& B = *m_BMatrix;
-      M& Wy = *m_observationErrorMatrix;
```
roystgnr, line 1238: Weren't we saving both the precision and the error forms of this matrix now?

roystgnr:
We'll need to cherry pick it into a new branch, of course, and we need to go back later to implement this in a better way, but otherwise the second commit here looks fine.

roystgnr:
Actually, wait.  How did that second commit only affect the test_gpmsa_autoscaled_samples.m regression solution, and not test_gpmsa_vector_samples.m too?

roystgnr:
Closing in favor of #643, which has a more efficient version of the Wy fix and isn't based off a broken PR.

PR #639, opened 2017-09-25
Fix bug in K setup

dmcdougall:
We're taking eigenvectors from the wrong place.  We now take them from the
truncated SVD list, as they're in the correct order.

This PR replaces #635.  PR #635 implemented the incorrect fix.


roystgnr:
I take it this fix also worked for you on the small case?  Have you tried it on the larger case yet?

dmcdougall:
We now match (up to sign) Brian's K matrix.  I'm not sure how this affects the log pdf values because I'm relying on other fixes in other branches I didn't want to put all in this pull request.

Issue #569, opened 2017-04-21
Mixed input file and C++ API options handling


briadam:
Enhance other (Environment, Metropolis Hastings, inverse problem) constructors with options setting flexibility similar to GPMSA, so one can default construct, set via C++ API, or parse from input file in flexible order.  In particular, it's not possible to set through C++, then override with power user options from an input file.

See discussion with @roystgnr https://groups.google.com/forum/#!topic/queso-users/S58W-IzRRtI

Related, there is a potential chicken and egg problem with the options for constructing an Environment, as parsing Environment options from a file requires an Environment...

dmcdougall:
If the user passes an options object, [the input file never gets parsed](https://github.com/libqueso/queso/blob/v0.57.0-release/src/core/src/Environment.C#L1230-L1249).  Probably want to mess around with that logic.

briadam:
@dmcdougall @roystgnr  For the Environment (and other) constructors that accept both input file name and EnvOptionsValues object, what do you feel the precedence should be?  

On the Dakota side, we're planning to have the C++ options object apply first, followed by the parsed input file options as this lets a user write a C++ program with their preferred defaults possibly overriding QUESO class defaults, but then lets power users override them at run-time through a file.  So our thinking is (1) class default values, possibly overridden by (2) explicitly set options in C++ options object, possibly overridden by (3) user input file.

briadam:
Per @dmcdougall, also need to consider where in this workflow a client might programmaticly set options as well, e.g., to override defaults through C++ code.

roystgnr:
Although the goal is to give the user as much control over the ordering as possible, there's no way to do that when you have two options sources going into the same constructor, and "C++ object gets used first, but input file can override anything in it" is clearly the way to go.  We can make sure there's accessors available if anyone wants to then override an input file option from C++ even later.

briadam:
Where's the right place in the code base to put a convenience function for converting sets of scalars to a string as I am doing in EnvironmentOptions for default management?  
Edit: Never mind, I went with Miscellaneous.h/C

briadam:
Should I leave SequenceStatisticalOptions in place or remove it?  Or at least remove its use in other Options classes?  I see it's deprecated and it's complicating MetropolisHastingsSGOptions among other Options classes...

briadam:
Related, should other deprecated Options classes like MetropolisHastingsSGOptions be left in place or removed for this refactor?

roystgnr:
[...]

PR #640, opened 2017-09-25
More granular options parsing for Environment

briadam:
I'm seeking preliminary feedback on this before forging ahead.

Tests break when Boost Program Options are enabled, but they seem to on dev as well.  Guidance?
  FAIL: test_BoostInputOptionsParser
  FAIL: t01_valid_cycle/rtest01.sh
  (HANGS or SLOW): test_mala

I'm going to mark up the code myself with questions.  Also these general ones:
 * Should I further condense code shared between FullEnvironment::construct(...) variants?
 * Do we anticipate a need for options_have_been_used checking as in GPMSA?

### DIFF
--- a/src/core/inc/EnvironmentOptions.h
+++ b/src/core/inc/EnvironmentOptions.h
@@ -94,6 +94,15 @@ public:
   //! Copy constructor
   EnvOptionsValues(const EnvOptionsValues& src);
 
+  //! Set parameter option names to begin with prefix
+  void set_prefix(const char* prefix);
+
+  //! Set default values for parameter options
+  void set_defaults();
+
+  //! Given prefix, read the input file for parameters named "prefix"+*
+  void parse(const BaseEnvironment* env, const char* prefix);
+
   //! Destructor
   virtual ~EnvOptionsValues();
   //@}
--- a/src/core/inc/EnvironmentOptions.h
+++ b/src/core/inc/EnvironmentOptions.h
@@ -95,13 +95,13 @@ public:
   EnvOptionsValues(const EnvOptionsValues& src);
 
   //! Set parameter option names to begin with prefix
-  void set_prefix(const char* prefix);
+  void set_prefix(const std::string& prefix);
 
   //! Set default values for parameter options
   void set_defaults();
 
   //! Given prefix, read the input file for parameters named "prefix"+*
-  void parse(const BaseEnvironment* env, const char* prefix);
+  void parse(const BaseEnvironment& env, const std::string& prefix);
 
   //! Destructor
   virtual ~EnvOptionsValues();
--- a/src/core/src/Environment.C
+++ b/src/core/src/Environment.C
@@ -150,7 +150,7 @@ FilePtrSetStruct::~FilePtrSetStruct()
 // Default constructor --------------------------------
 BaseEnvironment::BaseEnvironment(
   const char*                    passedOptionsInputFileName,
-  EnvOptionsValues* alternativeOptionsValues)
+  EnvOptionsValues* initialOptionsValues)
   :
   m_fullEnvIsReady             (false),
   m_worldRank                  (-1),
@@ -182,14 +182,19 @@ BaseEnvironment::BaseEnvironment(
   // If the user passed in an options object pointer, we really shouldn't let
   // ScopedPtr delete their object, so we make a copy.  That way, the dtor
   // will kill this local copy and leave the user's object in tact.
-  if (alternativeOptionsValues != NULL) {
-    m_optionsObj.reset(new EnvOptionsValues(*alternativeOptionsValues));
+  if (initialOptionsValues != NULL) {
+    m_optionsObj.reset(new EnvOptionsValues(*initialOptionsValues));
+  }
+  else {
+    // BMA 20170925: Changed this to default construct the options when none
+    // are passed in, so downstream checks for != NULL may behave differently
+    m_optionsObj.reset(new EnvOptionsValues());
   }
 }
 
 BaseEnvironment::BaseEnvironment(
   const std::string&             passedOptionsInputFileName,
-  EnvOptionsValues* alternativeOptionsValues)
+  EnvOptionsValues* initialOptionsValues)
   :
   m_fullEnvIsReady             (false),
   m_worldRank                  (-1),
@@ -219,8 +224,13 @@ BaseEnvironment::BaseEnvironment(
   // If the user passed in an options object pointer, we really shouldn't let
   // ScopedPtr delete their object, so we make a copy.  That way, the dtor
   // will kill this local copy and leave the user's object in tact.
-  if (alternativeOptionsValues != NULL) {
-    m_optionsObj.reset(new EnvOptionsValues(*alternativeOptionsValues));
+  if (initialOptionsValues != NULL) {
+    m_optionsObj.reset(new EnvOptionsValues(*initialOptionsValues));
+  }
+  else {
+    // BMA 20170925: Changed this to default construct the options when none
+    // are passed in, so downstream checks for != NULL may behave differently
+    m_optionsObj.reset(new EnvOptionsValues());
   }
 }
 
@@ -1200,9 +1210,9 @@ FullEnvironment::FullEnvironment(
   RawType_MPI_Comm             inputComm,
   const char*                    passedOptionsInputFileName,
   const char*                    prefix,
-  EnvOptionsValues* alternativeOptionsValues)
+  EnvOptionsValues* initialOptionsValues)
   :
-  BaseEnvironment(passedOptionsInputFileName,alternativeOptionsValues)
+  BaseEnvironment(passedOptionsInputFileName,initialOptionsValues)
 {
   this->construct(inputComm, prefix);
 }
@@ -1211,9 +1221,9 @@ FullEnvironment::FullEnvironment(
   RawType_MPI_Comm             inputComm,
   const std::string&           passedOptionsInputFileName,
   const std::string&           prefix,
-  EnvOptionsValues* alternativeOptionsValues)
+  EnvOptionsValues* initialOptionsValues)
   :
-  BaseEnvironment(passedOptionsInputFileName,alternativeOptionsValues)
+  BaseEnvironment(passedOptionsInputFileName,initialOptionsValues)
 {
   this->construct(inputComm, prefix.c_str());
 }
@@ -1247,26 +1257,9 @@ FullEnvironment::construct (RawType_MPI_Comm inputComm,
   std::cout << "In FullEnv, finished dealing with MPI initially" << std::endl;
 #endif
 
-  //////////////////////////////////////////////////
   // Read options
-  //////////////////////////////////////////////////
-  // If NULL, we create one
-  if (m_optionsObj == NULL) {
-    // If there's an input file, we grab the options from there.  Otherwise the
-    // defaults are used
-    if (m_optionsInputFileName != "") {
-#ifndef QUESO_DISABLE_BOOST_PROGRAM_OPTIONS
-      m_allOptionsMap.reset(new boost::program_options::variables_map());
-      m_allOptionsDesc.reset(new boost::program_options::options_description("Allowed options"));
-#endif  // QUESO_DISABLE_BOOST_PROGRAM_OPTIONS
-
-      readOptionsInputFile();
-
-      m_input->parse_input_file(m_optionsInputFileName);
-    }
-
-    m_optionsObj.reset(new EnvOptionsValues(this, prefix));
-  }
+  if (!m_optionsInputFileName.empty())
+    readOptionsInputFile(prefix);
 
   // If help option was supplied, print info
   if (m_optionsObj->m_help != "") {
@@ -1491,9 +1484,9 @@ FullEnvironment::construct (RawType_MPI_Comm inputComm,
 FullEnvironment::FullEnvironment(
   const char*                    passedOptionsInputFileName,
   const char*                    prefix,
-  EnvOptionsValues* alternativeOptionsValues)
+  EnvOptionsValues* initialOptionsValues)
   :
-  BaseEnvironment(passedOptionsInputFileName,alternativeOptionsValues)
+  BaseEnvironment(passedOptionsInputFileName,initialOptionsValues)
 {
   this->construct(prefix);
 }
@@ -1501,9 +1494,9 @@ FullEnvironment::FullEnvironment(
 FullEnvironment::FullEnvironment(
   const std::string&             passedOptionsInputFileName,
   const std::string&             prefix,
-  EnvOptionsValues* alternativeOptionsValues)
+  EnvOptionsValues* initialOptionsValues)
   :
-  BaseEnvironment(passedOptionsInputFileName,alternativeOptionsValues)
+  BaseEnvironment(passedOptionsInputFileName,initialOptionsValues)
 {
   this->construct(prefix.c_str());
 }
@@ -1511,6 +1504,7 @@ FullEnvironment::FullEnvironment(
 void
 FullEnvironment::construct (const char *prefix)
 {
+
 #ifdef QUESO_MEMORY_DEBUGGING
   std::cout << "Entering FullEnv" << std::endl;
 #endif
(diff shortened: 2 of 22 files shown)


src/core/src/Environment.C
```diff
@@ -219,8 +224,13 @@ BaseEnvironment::BaseEnvironment(
...
   // will kill this local copy and leave the user's object in tact.
-  if (alternativeOptionsValues != NULL) {
-    m_optionsObj.reset(new EnvOptionsValues(*alternativeOptionsValues));
+  if (initialOptionsValues != NULL) {
+    m_optionsObj.reset(new EnvOptionsValues(*initialOptionsValues));
+  }
+  else {
+    // BMA 20170925: Changed this to default construct the options when none
+    // are passed in, so downstream checks for != NULL may behave differently
+    m_optionsObj.reset(new EnvOptionsValues());
```
briadam, line 233: I'm now default constructing EnvOptionsValues if not passed in.  Is that a problem?  Should I remove checks such as:
  if (m_optionsObj == NULL) return ".";
  queso_require_msg(m_optionsObj, "m_optionsObj variable is NULL");

  What about for EmptyEnvironment?  Is it important that it have options = NULL as opposed to just default constructed options?  The only use case is for member BaseTKGroup::m_emptyEnv which doesn't appear to get used anywhere...

src/core/inc/EnvironmentOptions.h
```diff
@@ -94,6 +94,15 @@ class EnvOptionsValues
...
   EnvOptionsValues(const EnvOptionsValues& src);
 
+  //! Set parameter option names to begin with prefix
+  void set_prefix(const char* prefix);
+
+  //! Set default values for parameter options
+  void set_defaults();
+
+  //! Given prefix, read the input file for parameters named "prefix"+*
+  void parse(const BaseEnvironment* env, const char* prefix);
```
briadam: I kept pointers such as const char* and Env* for consistency with ctor instead of changing APIs to use std::string and Env&

src/core/src/EnvironmentOptions.C
```diff
@@ -46,139 +46,24 @@ namespace QUESO {
...
-    m_seed(UQ_ENV_SEED_ODV),
-    m_platformName(UQ_ENV_PLATFORM_NAME_ODV),
-    m_identifyingString(UQ_ENV_IDENTIFYING_STRING_ODV),
-    m_numDebugParams(UQ_ENV_NUM_DEBUG_PARAMS_ODV),
-    m_debugParams(m_numDebugParams,0.),
-    m_env(NULL),
+  m_env(NULL)
 #ifndef QUESO_DISABLE_BOOST_PROGRAM_OPTIONS
-    m_parser(),
+  , m_parser(new BoostInputOptionsParser())
```
briadam, line 51: Why do we default construct a BoostInputOptionsParser for the empty input file case?

src/core/src/EnvironmentOptions.C
```diff
@@ -264,4 +149,174 @@ std::ostream& operator<<(std::ostream& os, const EnvOptionsValues & obj)
...
+  m_subDisplayAllowInter0 = UQ_ENV_SUB_DISPLAY_ALLOW_INTER0_ODV;
+  //m_subDisplayAllowedSet  = ;
+  m_displayVerbosity = UQ_ENV_DISPLAY_VERBOSITY_ODV;
+  m_syncVerbosity = UQ_ENV_SYNC_VERBOSITY_ODV;
+  m_checkingLevel = UQ_ENV_CHECKING_LEVEL_ODV;
+  m_rngType = UQ_ENV_RNG_TYPE_ODV;
+  m_seed = UQ_ENV_SEED_ODV;
+  m_platformName = UQ_ENV_PLATFORM_NAME_ODV;
+  m_identifyingString = UQ_ENV_IDENTIFYING_STRING_ODV;
+  m_numDebugParams = UQ_ENV_NUM_DEBUG_PARAMS_ODV;
```
briadam, line 169: Why aren't debug parameters parsed from the input file?  Should I add them?

src/core/src/EnvironmentOptions.C
```diff
@@ -264,4 +149,174 @@ std::ostream& operator<<(std::ostream& os, const EnvOptionsValues & obj)
...
+    "output filename for subscreen writing");
+  m_parser->registerOption<bool>
+    (m_option_subDisplayAllowAll, m_subDisplayAllowAll,
+    "Allow all processors to write to output file");
+  m_parser->registerOption<bool>
+    (m_option_subDisplayAllowInter0, m_subDisplayAllowInter0,
+    "Allow all inter0 nodes to write to output file");
+
+  // convert the current member set of integers to a string for default handling
+  std::ostringstream sdas_as_string;
```
briadam: I'm converting the set<unsigned int> back to string for default handling.  Is the trailing whitespace a problem?

src/core/src/EnvironmentOptions.C
```diff
@@ -264,4 +149,174 @@ std::ostream& operator<<(std::ostream& os, const EnvOptionsValues & obj)
...
+
+}
+
+
+void
+EnvOptionsValues::parse(const BaseEnvironment* env, const char* prefix)
+{
+  m_env = env;
+
+  if (m_env->optionsInputFileName().empty()) {
```
briadam: Should parse be robust to the case of an empty input file string?  For now, I'm leaving the error in place.  Seems calling parse means you really want to parse, but I could also see just silently skipping the parse if nothing to do.

briadam (commented):
Questions inlined for @dmcdougall  and @roystgnr ...

src/core/src/Environment.C
roystgnr: We're going with "file options override constructor-time C++ options, right?  In that case default-constructing EnvOptionsValues seems like the right thing to do; it gives us a copy of the default options, unless they've been overridden in C++, and either way they're ready for the "input files can override us still further" code path.

roystgnr: The ```return ".";``` does seem redundant now, since that's the default value for ```EnvOptionsValues::m_subDisplayFileName```.

roystgnr: I'd leave the queso_require in there.  Since it's not in user hands anymore we could turn it into a queso_assert if you like, but it's also not in an inner loop anymore so there's no compelling reason to.

roystgnr: And there seems no reason not to give EmptyEnvironment default options, to make it less useless.

src/core/src/Environment.C
```diff
@@ -1247,26 +1257,9 @@ FullEnvironment::construct (RawType_MPI_Comm inputComm,
...
-
-      readOptionsInputFile();
-
-      m_input->parse_input_file(m_optionsInputFileName);
-    }
-
-    m_optionsObj.reset(new EnvOptionsValues(this, prefix));
-  }
+  if (!m_optionsInputFileName.empty())
+    readOptionsInputFile(prefix);
```
roystgnr, line 1264: This is a semantic change, not just an improved capability, so let's be wary of regressions.  I can't imagine anyone was actually *using* the behavior of "pass an input file name and a non-null option object pointer in and expect the former to be ignored", but I've seen stranger code in the past.

src/core/inc/EnvironmentOptions.h
roystgnr: I wouldn't worry about consistency in a brand-new API, and I wouldn't worry too much about stylistic consistency in QUESO.

Using a reference is IMHO always better than a pointer to a class when you expect the pointer to never be null.

Using a char pointer rather than converting to std::string is often a nice way to buy a bit of efficiency at the expense of a bit of safety, but the former is negligible in a method that'll only be called a handful of times at most per execution, so I'd prefer the latter.

src/core/src/Environment.C
```diff
@@ -1491,26 +1484,27 @@ FullEnvironment::construct (RawType_MPI_Comm inputComm,
 FullEnvironment::FullEnvironment(
   const char*                    passedOptionsInputFileName,
   const char*                    prefix,
-  EnvOptionsValues* alternativeOptionsValues)
+  EnvOptionsValues* initialOptionsValues)
```
roystgnr, line 1489: I like the name changes.  "Self-documenting code" rather than "atavistic deceptive code" is a big plus.

src/core/src/Environment.C
```diff
@@ -1782,9 +1759,17 @@ void queso_terminate_handler()
 
 
 //-------------------------------------------------------
+// Previous behavior:
+//   * IF      alternative options passed in, use them exclusively
+//   * ELSE IF input file, parse and set options from it
+//   * ELSE    use default options
+// New behavior:
+//   * START   with default options or passed alternative options
+//   * IF      input file, any parsed options override stored values
```
roystgnr: I'd put this comment in the header, where many people will read it as they use the API, rather than in the source file, where nobody but core developers will look unless they're already tearing their hair out over failure to understand the behavior.

src/core/src/EnvironmentOptions.C
roystgnr: Wait, isn't the ```new BoostInputOptionsParser()``` bit just in your patch?  The previous code was just ```m_parser()```, leaving it an uninitialized scoped pointer.

src/core/src/EnvironmentOptions.C
roystgnr: These parameters appear to be an atavism that haven't actually been *used* for anything:
* since 2010, at the latest
* in the version control master, rather than a branch
* in git, rather than svn

So our second best option here would be "dig through ancient svn branches to figure out what these are for", and our best would be "just get rid of them".

src/core/src/EnvironmentOptions.C
roystgnr: Nah, keep the error.  Belt-and-suspenders.

src/core/src/EnvironmentOptions.C
roystgnr: Actually, I'm pretty sure it is.  Whitespace in the input *files* is fine with either boost or getpot, but I'm not sure either trims whitespace from input *API parameters*, and if they do it's surely luck rather than officially supported behavior.

roystgnr:
> Tests break when Boost Program Options are enabled, but they seem to on dev as well. Guidance?
> FAIL: test_BoostInputOptionsParser
> FAIL: t01_valid_cycle/rtest01.sh
> (HANGS or SLOW): test_mala

I just discovered that my gpmsa_new_functional branch doesn't even *link* when boost::program_options is enabled, so I'm merely embarrassed rather than surprised that dev isn't getting much love for that configuration either.  I could swear I had it working in January...

roystgnr:
test_mala finished quickly for me, but the log file is filled with NaN warnings.

roystgnr:
The other two failed.

roystgnr:
```
  what():  the argument (''false'') for option 'bool' is invalid. Valid choices are 'on|off', 'yes|no', '1|0' and 'true|false'
```

Coincidentally, "what" was my reaction as well.

src/core/src/EnvironmentOptions.C
briadam: Yes, you're right.  I caused confusion here by not making reference and relaying a question I was more asking myself.  I took a cue from https://github.com/libqueso/queso/blob/c6c71a0ee2bc81c7316fe67dbff104dc8104b64e/src/gp/src/GPMSAOptions.C#L133 where it gets initialized, but I agree leaving the scoped ptr uninitialized should be okay.

src/core/src/EnvironmentOptions.C
briadam: Okay, if @dmcdougall agrees, I can purge them.

roystgnr:
I saw Goody @dmcdougall [dancing with the devil](457)!

GetPot appears to be much more flexible when in how it deals with booleans and a bit trickier with strings and non-scalar options - it strips enclosing quotes if they're there, while boost::program_options gets confused by quotes around booleans and retains quotes around strings and non-scalar values.

So I think we'd actually be fine if we just used boost-compatible files for everything (and reverting the single quotes added to test01_valid_cycle/tgaCycle.inp backs me up on this)... which makes me wonder why exactly Damon went to all the trouble of changing them in the first place.

roystgnr:
> Should I further condense code shared between FullEnvironment::construct(...) variants?

On the one hand, [I'm a big fan of refactoring that reduces code size](https://www.folklore.org/StoryView.py?story=Negative_2000_Lines_Of_Code.txt), but on the other hand only you know [if that's worth the time](https://quoteinvestigator.com/2012/04/28/shorter-letter/).

roystgnr:
> Do we anticipate a need for options_have_been_used checking as in GPMSA?

No, but the whole point of that checking was to handle unanticipated application code mistakes, so it wouldn't hurt to use that idiom more often, if you want.

src/core/src/EnvironmentOptions.C
roystgnr: Ah, that looks like a mistake on my part.  If we need m_parser then we reset it anyway, so there's no point in initializing it in the first place...

briadam:
Thanks for all the helpful review comments!
Yeah I see the same as you with t01_valid_cycle ([edit] and test_BoostInputOptionsParser).   I still have 1 fail to chase down. I'm going to focus on integrating your review first and then if you haven't picked this issue back up, can revisit.

briadam:
> So I think we'd actually be fine if we just used boost-compatible files for everything (and reverting the single quotes added to test01_valid_cycle/tgaCycle.inp backs me up on this)... which makes me wonder why exactly Damon went to all the trouble of changing them in the first place.

I made the rest of Roy's requested changes.  Should I make a commit on this branch (or dev) to reverse the input file changes from #457 ?  Doing so seems to fix most tests with BPO enabled, though not test_intercomm0, so there might be a real regression with my changes.

dmcdougall:
> So I think we'd actually be fine if we just used boost-compatible files for everything (and reverting the single quotes added to test01_valid_cycle/tgaCycle.inp backs me up on this)... which makes me wonder why exactly Damon went to all the trouble of changing them in the first place.

I could've sworn they weren't cross-compatible.  That's on me.

src/core/src/EnvironmentOptions.C
dmcdougall: I'm ok with this.

dmcdougall:
I'm ok with doing the file-change in a different PR.

roystgnr:
Isn't test_intercomm0 the one that we expect to randomly fail, because it doesn't use a fixed PRNG seed?  The price of stochastic regression testing...

roystgnr:
Yeah, let's do the input format reversion in a separate PR.

briadam:
Well good news, it (test_intercomm0) is behaving as expected then!  Issue #642 and PR #644 should resolve that part of this discussion.

briadam:
@dmcdougall I'm working to preserve BPO support in this PR, if that affects which release we target with this.  I realize other aspects of this commit do change behavior, but at @roystgnr pointed out, only in isolated ways.  I'm thinking to take a second pass to remove BPO support once you draw a release with the first pass options refactor.

dmcdougall:
> Isn't test_intercomm0 the one that we expect to randomly fail, because it doesn't use a fixed PRNG seed?

No.  It should pass every time.

briadam:
@dmcdougall @roystgnr I decided to make this change in a way that's largely backward compatible and preserves Boost Program Options parsing.  Would prefer this PR make it into the next release as it's needed for effective use of GPMSA scalar capability in Dakota.

roystgnr:
Can you fix up the conflicts?

briadam:
Yes, working on it now.  The Github docs suggested a process that would mean I make a merge commit on dev, which is surely not right.  Should I just merge dev into my options_handling_569 branch and resolve the conflict?  Or something else?

briadam:
Well, that'll teach me to try to do the merge with the webtool.  I'll do locally this afternoon and fix this up...

roystgnr:
I usually "git fetch origin" and "git rebase origin/dev" and then manually fix conflicts as they come up in the rebase.  That's sometimes easier said than done, though.

briadam:
Okay I was just being silly and had messed up that merge.  I rebased and force pushed to overwrite the bum merge.  I'm hoping this passes checks now.  

However, there's a pre-existing issue on dev that prevents building with Boost Program Options:  #655  Probably merge this PR and then see if anyone can fix that.  I struggled to get bisect working right.

briadam:
Assuming checks pass, this can be merged.  I'll note some cleanup needs in #569 and hopefully take one more pass before calling that done.

briadam:
I'm planing to take a cleanup pass today, including adding public default ctors to all the options classes I mucked with.

src/core/src/EnvironmentOptions.C
briadam: @roystgnr  FWIW, I learned the hard way why this initialization is necessary: operator<< dereferences this object, which would fail after default construction.  I added initialization to all classes for now and will promptly remove it post 0.58.0.  Same bug could affect copy ctors, but that use case isn't exercised by any code right now, so ignoring.

briadam:
@roystgnr I rebased this on top of your GPMSAOption BPO fix.  A quick review wouldn't hurt, but it's a ton of code, and my recent revisions shouldn't be controversial.  I'm good for merge when you are.

src/stats/inc/StatisticalForwardProblemOptions.h
```diff
@@ -109,7 +118,7 @@ class SfpOptionsValues
 
 private:
 #ifndef QUESO_DISABLE_BOOST_PROGRAM_OPTIONS
-  BoostInputOptionsParser * m_parser;
+  ScopedPtr<BoostInputOptionsParser>::Type m_parser;
```
roystgnr, line 121: Shoot, this fixed a memory leak, didn't it?

Thanks!

roystgnr (approved):

roystgnr:
Pinging @dmcdougall - IMHO this should be good to merge, but if you've got time it wouldn't hurt to get your eyes on it too first.

roystgnr:
@dmcdougall, in person: "I trust you."

I'm sure that'll come back to haunt him later but for now it just means we're ready to merge.

PR #643, opened 2017-09-29
Wy is precisions, the inverse of errors

roystgnr:
But we've already got a precision matrix handy, so let's use that
rather than recomputing it, @dmcdougall.

This should obsolete #636.


dmcdougall:
That feeling when you take too long to fix a pull request, so @roystgnr closes it and opens a new one with the correct implementation.  :p

Issue #642, opened 2017-09-28
Single quotes break Boost program options


briadam:
On the dev branch, using single quotes around booleans, strings, and sets breaks input file parsing when using Boost program options.  This is counter to the input file changes of @dmcdougall in #457  See the discussion in #640, where @roystgnr asserts that reverting #457 is likely okay.  

Specifically these tests break:
- FAIL: test_BoostInputOptionsParser
- FAIL: test_intercomm0/test_intercomm0_gravity_run.sh
- FAIL: t01_valid_cycle/rtest01.sh
- FAIL: test_mala

dmcdougall:
Does removing the quotes produce a file that is acceptable by both BPO and GetPot?

briadam:
Resolved by #644

PR #644, opened 2017-09-29
Revert quotes in input files that can foil Boost prog opts

briadam:
(Merging this will leave one failure in test_intercomm0, at least on my system, when BPO is enabled.  My understanding, subject to stray cosmic rays, is that this test is expected to be stochastic.)

Partially addresses libqueso/queso#642 by partially reverting
  "Make boost po opt-in, convert all test input files to getpot format"

This reverts the input files modified in
  commit 00477bdb79b3588999ef1298c69884fa0f42146e.

Conflicts:
	configure.ac


briadam:
Hmm, Travis reports test_uqGaussianVectorRVClass fails.  Which for me is a stochastic fail that I haven't seen in a while.  Is this expected?

briadam:
Well it appears the seed for that test is taken from /dev/random, so I guess unlikely that the run would be identical, but the stochsatic test appears to have a tolerance such that it should pass cross-platform?

roystgnr:
Sadly, there is no such thing as "a tolerance such that it should pass cross-platform", only "a tolerance such that it should pass cross-platform N% of the time".  Hence the pass with clang and failure with gcc this time.

I don't know offhand how to "kick" a Travis check other than by doing an interactive rebase, minor commit message change, and force push.  You can try that if you like.

I'll build and test this branch on a couple of my own machines, and if it passes on both then I'll merge.

roystgnr (approved):

briadam:
This worked to restart it, once I signed into Travis using my GitHub account: https://stackoverflow.com/questions/17606874/trigger-a-travis-ci-rebuild-without-pushing-a-commit

roystgnr:
There might actually be something seriously wrong here.  The failure seems outright deterministic for me on one system.
```
$ wc $(find -name sfp_gravity_p_seq.m)
  50100   50300 1203450 ./test/output_test_intercomm0_gravity_2/sfp_gravity_p_seq.m
  51102   51306 1227519 ./test/output_test_intercomm0_gravity_1/sfp_gravity_p_seq.m
```

They mostly match in the diff, but the latter file has a thousand samples which the former file doesn't.

What's especially weird is that they have TEN TIMES as many samples as the same runs on a different system, and neither matches what I'd have expected to see - doesn't the input file ask for ip_mh_rawChain_size of 2e4, not 5e3 or 5e4??

briadam:
Yes, I saw that behavior as well, though I think I saw it with test_intercomm0, not test_uqGaussianVectorRVClass.  The latter passes in all my local builds.  I also thought something is actually wrong, given that it consistently PASSed with GetPot and FAILed with BPO.  However, I stopped digging when you told me it was an inconsistent regression test.  I'm probably not qualified to root cause it, but can attempt to dig back in if needed, especially given some pointers on where to look.

roystgnr:
Sorry, didn't mean for any confusion, it actually *is* test_intercomm0 that's failing (on one system) for me.  test_uqGaussianVectorRVClass is working fine with every test configuration I throw at it.

briadam:
Not sure what your team development practices are w.r.t. this, but it seems that test_intercomm0 w/BPO is a pre-existing instability on dev and not affected by this PR, so we could merge it.  Would you rather track and resolve in this issue #642 / PR #644, #640, or should I create a new issue?  I know it sucks to have to keep track of second-order failures like "this PR doesn't make the existing failures any worse," so I'm happy to handle however you like.

roystgnr:
Yeah, test_intercomm0 is a failure on dev for me too.  (Although again, only on one of the systems I tried...)  I'll try bisecting it (although I swear it was working for me on the gpmsa_new_functional branch, which is based on dev!) and swapping around MPI stacks and see what I can figure out, but IMHO there's no reason it should hold up this PR.

dmcdougall:
> My understanding, subject to stray cosmic rays, is that this test is expected to be stochastic.

No, test_intercomm0 should pass every time.

dmcdougall:
> They mostly match in the diff, but the latter file has a thousand samples which the former file doesn't.

Chain size is 20000 and filtered chain lag is 20.  20000 / 20 = 1000.

> What's especially weird is that they have TEN TIMES as many samples as the same runs on a different system, and neither matches what I'd have expected to see

QUESO appends output to output files, not truncate.  I suspect this clearly broken `rm` command is the cause of the problem.

dmcdougall:
Oops, this is the url I meant to paste in my previous comment: https://github.com/libqueso/queso/blob/dev/test/test_intercomm0/test_intercomm0_gravity_run.sh#L6

dmcdougall:
Disregard above.

PR #645, opened 2017-09-29
Fix test_intercomm0 false positives

roystgnr:
The executables here are each *appending* to their output files, which
means that any one failure turns into a failure forevermore, or at
least until the build dir is wiped and recreated.

Instead deleting and recreating output files for each test run fixes
the problem.


roystgnr:
Hopefully this fixes the bug for you, too, @briadam?

briadam (approved):
Yeah, we get burned on the Dakota side with stale output too, but are loathe to remove files Dakota didn't create, or more typically, are present from an earlier Dakota run.  In this case, though fix looks good to me.

When applied to head of dev, this fixes the issue for me on RHEL7, both for GetPot and BPO builds.

PR #648, opened 2017-10-04
GPMSA: Correct latent indexing bug for autoscale

briadam:
In two tests and one example, autoscale was getting set on a
per-simulation basis, not for each parameter, scenario, or response
index.

It likely worked since numSimulations > numIndices for each of those.


roystgnr:
Yikes; good find.

roystgnr (approved):

PR #649, opened 2017-10-05
 Retire deprecated Options classes

briadam:
Remove classes SequenceStatisticalOptions and
MetropolisHastingsSGOptions.  Required writing new copy constructor so
MhOptionsValues could be built from an MLSampleLevelOptions object.

This to be merged to dev prior to completion of #569 I don't think there's anything too controversial to request review on, but wouldn't mind a quick look.  Some of the code will cleanup as I finish up #569


briadam:
Actually one outstanding question: can code protected by 
```
#ifdef QUESO_USES_SEQUENCE_STATISTICAL_OPTIONS
```
be removed?  SsOptionsValues in particular?  That preprocessor define is hardwired off in Defines.h, but associated code is not marked deprecated, so I left it in place, although didn't test as flipping that define caused all manner of chaos.

roystgnr:
That define looks like it's been hardwired since 2012, and we're obviously not supporting it, so I'd say strip it out, but let's wait to hear from @dmcdougall to confirm first.

roystgnr (approved):
I think I'd have factored into commits slightly differently, but the end result looks good and the intermediate commit passes "make check" too so I'm happy.

briadam:
Yes, I confirmed it passed in between.  I would have done differently as well, but had missed some code on my first pass (due to conditional compilation) and it didn't seem worth a careful revert and re-apply.  I'm thinking I'll merge this PR and then pick up the sequence stat options based on Damon's feedback.

Issue #647, opened 2017-10-04
Inconsistent includes refer to */inc/*.h instead of queso/*.h


briadam:
As reported by @lpswile several .C files (listed below) refer to files as */inc/*.h.  I'm hoping these can instead be queso/*.h for consistency at build vs. install time.  @roystgnr @dmcdougall Is it appropriate to correct them?

```
egrep -R '^#include.+/inc/' * 
src/stats/src/MetropolisHastingsSG.C:#include "core/inc/FilePtr.h"
src/stats/src/MLSampling.C:#include "core/inc/FilePtr.h"
src/stats/src/MLSampling.C:#include "stats/inc/FiniteDistribution.h"
src/stats/src/SampledScalarCdf.C:#include "core/inc/FilePtr.h"
src/stats/src/StatisticalForwardProblem.C:#include "core/inc/FilePtr.h"
src/stats/src/MonteCarloSG.C:#include "core/inc/FilePtr.h"
src/stats/src/FiniteDistribution.C:#include "stats/inc/FiniteDistribution.h"
src/core/src/GslMatrix.C:#include "core/inc/FilePtr.h"
src/core/src/TeuchosVector.C:#include "core/inc/FilePtr.h"
src/core/src/GslVector.C:#include "core/inc/FilePtr.h"
src/core/src/TeuchosMatrix.C:#include <core/inc/FilePtr.h>
src/core/src/Environment.C:#include "core/inc/FilePtr.h"
src/basic/src/SequenceOfVectors.C:#include "core/inc/FilePtr.h"
src/basic/src/DiscreteSubset.C:#include "basic/inc/DiscreteSubset.h"
src/basic/src/ScalarSequence.C:#include "core/inc/FilePtr.h"
```

briadam:
Or perhaps this is intended as a build-time, only included in .C files vs. needed by a library client after installing kind of distinction.  If so, that's okay, but we're going to have to make the Dakota CMake build system smarter to treat that case.  Ugh.

roystgnr:
I'm afraid you've hit upon it.  Those three headers are listed as "Not installed" in Makefile.am.

Personally I like the idea of having "internal use only" headers whose APIs we can juggle without worrying about backwards compatibility, but IMHO if this is going to cause any delay with Dakota integration then we should just install the headers as a stopgap.

briadam:
[...]

PR #650, opened 2017-10-05
Consistently use queso/*.h for included headers

briadam:
Refer to included headers as queso/*.h instead of */inc/*.h

Addresses libqueso/queso#647 in a stopgap way.  While we'd prefer to
distinguish internal build-only headers from API headers, there are only
a few referred to as */inc/*.h.


roystgnr:
Did you add these headers to inc/queso/Makefile.am?  We'll still want to leave them out of libqueso_include_HEADERS in src/Makefile.am, so they don't get installed, but if they're going to be referred to as queso/Foo.h then we'll need to make those temporary symlinks in in the build directory.

briadam:
Nope and have no idea why my build worked.  Must have stale files around.  I'll fix.

PR #651, opened 2017-10-05
Consistently use queso/*.h for included headers

briadam:
Replaces PR #650 due to rebased branch

Refer to included headers as queso/*.h instead of */inc/*.h

Addresses libqueso/queso#647 in a stopgap way.  While we'd prefer to
distinguish internal build-only headers from API headers, there are only
a few referred to as */inc/*.h.

Adjust Makefile.am for build and install time header changes


roystgnr:
So, #650 actually went through fine.

Let me guess, you had the web interface open and the post-rebase #650 looked crazy to you from that?  That's actually a Github bug/limitation IIRC - it tries to only append to the comments page, and if you rewrite history then you're not going to see the canonical page until you reopen or refresh that page.

roystgnr (approved):
Looks great to me; let's let @dmcdougall have a chance at it before merging though.

roystgnr:
7 days ago?  He had his chance.

PR #653, opened 2017-10-11
Fix truncation error precision adjustments

dmcdougall:
We adjust the truncation error precision, not the emulator precision.
We also fix the adjustment to the scale parameter of the Gamma
distribution.  The adjustment in the paper is additive on the rate, not the
scale.

This commit also also updates the functional regression test solutions.


roystgnr:
This didn't change any of the multivariate regression solutions?

roystgnr (approved):
Good catches, on all of that.

dmcdougall:
> This didn't change any of the multivariate regression solutions?

You mean the multivariate regression solutions we won't have until we merge #654?  Nope.

:)

dmcdougall:
Oh I'm an idiot.

No, it didn't change the multivariate regressions.  That's surprising...

roystgnr:
Surprising but I suppose not inconceivable?  In Brian's email at least one of the corresponding changes only had an effect several decimal places out...

PR #654, opened 2017-10-11
GPMSA multivariate pdf test

dmcdougall:
This pull request depends on #653.  Please merge #653 before merging this pull request.


roystgnr:
Looks like a stochastic false positive on the check failure.

roystgnr:
And more tests (which we can even run and expect to pass, now!) are always good.  Thanks!

PR #656, opened 2017-10-16
GPMSAOptions boost codepath fixes

roystgnr:
This gets us compiling (and passing "make check", so at least the other options' parsing appears to be working) with --enable-boost-program-options.

I had to disable parsing of periodicity options for functional GPMSA discrepancy kernels, because I can't figure out *any* way other than manual string parsing to get a ```vector<bool>``` out of boost::program_options.  However, we don't have periodicity capabilities enabled yet, and by the time we do we ought to have stripped out the deprecated boost code paths.


briadam (approved):
Seems like a fine workaround for now, per our discussion.

Issue #658, opened 2018-02-21
Double-check units on the "covariance is too big" warning


dmcdougall:

briadam:
The conditional logic for this warning 
https://github.com/libqueso/queso/blob/469911ac41a8a829b994ac296e341386589a0721/src/stats/src/TransformedScaledCovMatrixTKGroup.C#L330-L343
is comparing units of variance (covariance diagonal) to domain bounds. Should probably instead by comparing sqrt(var) to the bounds.

PR #659, opened 2018-02-23
Correct units for "uniform-like proposal" warn

briadam:
Address libqueso/queso#658 by using sqrt(Covariance(i,i)) to compare
standard deviation to to domain bounds.


roystgnr:
:+1:

PR #664, opened 2019-05-06
Verify surrogate evaluation point

tradowsk:
Adds a helper function to `InterpolationSurrogateBase` to make sure each component of the surrogate evaluation point is within the proper bounds and is not NaN.

This was not checked before. When the `domainVector` was outside the surrogate bounds, it would just return garbage values and lead to difficult to diagnose issues upstream.

Also, what's a "surrogoate"? ;-p

### DIFF
--- a/src/surrogates/inc/InterpolationSurrogateBase.h
+++ b/src/surrogates/inc/InterpolationSurrogateBase.h
@@ -27,6 +27,9 @@
 
 #include <queso/SurrogateBase.h>
 
+#include <vector>
+
+
 namespace QUESO
 {
   // Forward declarations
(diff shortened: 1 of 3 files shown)


roystgnr:
This looks reasonable; I have no idea why gcov is screaming like mad.  I'd like it if we could fix that and rerun CI before merging but I'm not sure how.  I'd suggest just turning off the coverage stuff in .travis.yml for now, I'm afraid.

> Also, what's a "surrogoate"? ;-p

Those are [things that listen to nothing but calls and cries: Deaf, dumb, and blind, they are void of wisdom.](https://quran.com/2/171?translations=22).  Or is that just a surah-goat?  Quite an excellent metaphor for programming, anyway.

tradowsk:
This isn't a critical fix so there's no rush to merge.

I do have some Experimental Design code that will be ready for merge in a few weeks, and this would be good to have in there before that since the exp design code is what originally tripped this issue.

pbauman:
It looks like the Travis failure is the CodeCov failure, which seems unrelated to this.

PR #665, opened 2019-05-23
Discrete Experimental Design

tradowsk:
The purpose of this new infrastructure is to automate a sequential Bayesian experimental design step for inferring unknown model parameters from noisy experimental data. The goal is to identify the experimental scenario that is most informative to the current parameter belief distribution(s) based on the given metric. We iterate over every possible experimental scenario in the discrete scenario space, solve an inverse problem, calculate the metric, and then select the scenario with the highest metric value at the end. This infrastructure is general, but a specific example application can be found [here](https://github.com/tradowsk/exp-design) for surface catalycity models.

`GridSearchExperimentalDesign` is the main driver class that holds the bounds of the scenario space and the discretization size.

`ScenarioRunner` runs an inverse problem for scenario parameters passed by GridSearchExperimentalDesign.

`ExperimentalLikelihoodInterface` provides an interface for a user-defined likelihood class to access the scenario parameters for the current scenario. `ExperimentalLikelihoodWrapper` is a simple helper class for internal use only.

`ExperimentMetricBase` can be subclassed to allow for solving the inverse problem with either Metropolis-Hastings or MLSampling depending on the problem and desired metric. `ExperimentMetricMinVariance` uses M-H and favors posteriors with lower variances, while `ExperimentMetricEIG` uses the Expected Information Gain from MLSampling as the metric.

PR #664 should be merged before this.


pbauman:
Legit failures on Travis here with make check.

tradowsk:
Okay, I committed the autogenerated Makefiles for the first 2 commits, and also rebased on #664 so hopefully we should be good now

pbauman:
Discussed with @tradowsk earlier that he will add a self-contained example to this PR and then we can merge.

roystgnr:
:+1:

tradowsk:
I added a `discreteExperimentalDesign` example based on a simple kinetics problem for inferring activation energies at different temperatures. The example `.C` provides an outline that can be easily modified for use in a more complex problem.

roystgnr:
Let's see if this passes CI once we rebase it using #666.  You know, now that I've made a deal with the dev branch.

